PR #1, opened 2016-09-23
global: first commit

david-caro:
* Simply package all the jsonschemas for ease of distribution.


setup.py
```diff
@@ -0,0 +1,100 @@
...
+
+    return version
+
+
+if __name__ == '__main__':
+    setup(
+        name='inspire-schemas',
+        license='GPLv2',
+        author='CERN',
+        author_email='info@inspirehep.net',
```
kaplun: should be `admin@inspirehep.net`.

scripts/version_manager.py
```diff
@@ -0,0 +1,340 @@
+#!/usr/bin/env python
```
kaplun: Uh? Is this part of cookiecutter? Have you based the whole repo on invenio-cookiecutter? If not why?

inspire_schemas/records/institutions.json
```diff
@@ -0,0 +1,204 @@
...
+        "name_variants": {
+            "items": {
+                "properties": {
+                    "source": {
+                        "description": "FIXME: Do we really care about the source?",
+                        "enum": [
+                            "DESY",
+                            "DESY_AFF",
+                            "ADS",
+                            "INSPIRE-SCHEMAS.
```
kaplun: broken JSON here. (kudos to github)

inspire_schemas/records/elements/field.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "properties": {
+        "scheme": {
+            "title": "INSPIRE-SCHEMAS.,
```
kaplun: broken json

inspire_schemas/records/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE-SCHEMAS; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Inspirehep schemas."""
```
kaplun: I don't think we need an `__init__.py` for a directory that contins only data.

scripts/version_manager.py
david-caro: This is not part of cookiecutter, this is a helper to generate the version of the packages from the commit history (semantic versioning). I think that we should be extra careful with the versioning of this package, and imo, leaving that to a human driven process will very easily lead to mistakes sooner or later.

This is a very similar version of a script I used on another project, it works by checking the 'sem-ver' headers on commit messages and the tags. I'll add to the docs if you agree with this procedure.

kaplun: So you have started from cookiecutter and added this. Correct? Shall we propose this in an RFC to cookiecutter?

inspire_schemas/records/__init__.py
kaplun: Interesting. I didn't know about that. You mean so that it can then be referenced via `pkg_resources` right?

scripts/version_manager.py
david-caro: Sure, though this will mean changing the way of work, as this means that each commit in the main history tread does change the version, 'grouping' the merged branches into the parent merge commit if there.

The script also generates the changelog, something like:

```
* MINOR/BUG 0.0.2 "David Caro <david@dcaro.es>"
    68d54137: global: schemas and basic packaging
    FIXED ISSUES: http://github.com/inspirehep/inspire-schemas/issue/123

* MINOR/BUG 0.0.1 "David Caro <david@dcaro.es>"
    78652298: global: first commit
```
Where the issue link is detected by the `closes #...` or `fixes #...` and the Minor/bug header changes with the `sem-ver' header to tag features too (so you can grep). This is just an example so the link there does not work ;)

inspire_schemas/records/institutions.json
david-caro: I messed up trying to be smart with sed... :imp:

inspire_schemas/records/__init__.py
david-caro: yep, though we can do some tricks like putting it somewhere else and then exporting that path through entry points and such, that will require modifying or writing a wrapper to the functions that do the file loading.

inspire_schemas/records/institutions.json
kaplun: ehehe

scripts/version_manager.py
kaplun: So this means one release per commit. OK, that's extreme. Can't we have the best of both worlds and introduce a specific tag in a commit to trigger creating a new version?

Also `kwalitee` has the nice feature of grouping bugs with bugs, new features with new features. This would allow for nicer readibility.

inspire_schemas/records/hep.json
```diff
@@ -0,0 +1,888 @@
+{
```
kaplun: Please, call it `literature.json` since `hep` is just an historical artifact.

kaplun: Also we should introduce versioning to the filenames.

e.g. `literature-0.0.1.json`, `author-0.0.1.json` and generate a new file everytime we want to patch it.

scripts/version_manager.py
david-caro: It's not one release per commit, it's one version per commit. Imo you should not have the same package version from two different commits, and you should never have to relay on somebody to make sure that's the case, this solves both. This does not mean that you have to deploy every version, you can deploy only on tagged commits (not sure though how that is done on travis), and the version in the tag always has precedence, though I see nothing wrong on not releasing all the versions (release 1.1.0 but not 1.0.1), specially if the package will not have to be sold (thus, not having to satisfy superficial things like having always consecutive versions, or not having a version number 13).

I'm not sure if having a non-historical ordered list of changes improves readability (for me it does not, though I agree that's subjective). I can try to group like that, but imo that will be less useful (maybe more nice, but less useful).

If you really don't see this change, I can split into another pr.

inspire_schemas/records/hep.json
david-caro: This might require changing some of the code on inspire-next (to accommodate the new names), I'll check.

jacquerie: Yeah, this is harder than it looks.

scripts/version_manager.py
david-caro: btw. one thing is the changelog and another the release notes ;)

inspire_schemas/records/hep.json
kaplun: Better now, so then we have a path for upgrade on production :+1:

inspire_schemas/records/__init__.py
kaplun: OK. I am not sure I am understanding. Do what your :heart: tells you!

scripts/version_manager.py
kaplun: > I can try to group like that, but imo that will be less useful (maybe more nice, but less useful).

The value is that, in principle, I would be quickly able to scan through the release-notes to see if a given bug has been squashed, or I can check the incompatible changes to see what would break in my client code.

inspire_schemas/records/hep.json
david-caro: I vote for doing that on the next pull request (avoid this one from getting even bigger)

scripts/version_manager.py
david-caro: With the semantic versioning, the incompatible changes forcefully bump the major version, so you can just pin the requires to the major version and then be sure that unless there's a bug that was not picked up by ci/tests, it will never break your client code.

I'll try to get the release notes generated too (grouping bugs, features and apibreaking changes) since the preavious release, but in the meantime, just checking the entries where the major version changes in the changelog gives you an easy-to-check way to track those incompatible changes, and a grep/search for the bug id will give you the version, commit, author and link to the change that introduced the fix for it. Of course, if the developer did specify it in the commit message (that's harder to automate, as there's no way to know if the change is linked to an issue or if it's not, maybe we can force to open an issue for each pull request to make sure that at least there's a paper trail too for each change...)

inspire_schemas/records/hep.json
kaplun: Yeah, what I mean, is before we deploy :wink:

david-caro:
I've added the 'releasenotes' command to the versioning script, it's a bit bare, but it generates now something similar to what you requested, for example:

```
New changes for version 0.1.0
=================================

API Breaking changes
--------------------


New features
------------
* FEATURE 0.1.0 "David Caro <david@dcaro.es>"
    a88a1edc: global: schemas and basic packaging


Bugfixes and minor changes
--------------------------
* MINOR/BUGFIX 0.0.1 "David Caro <david@dcaro.es>"
    78652298: global: first commit
```

Or with some api breaking changes:
```
New changes for version 1.1.0
=================================

API Breaking changes
--------------------
* API_BREAKING/MAJOR 1.0.0 "David Caro <david@dcaro.es>"
    78652298: global: first commit


New features
------------
* FEATURE 1.1.0 "David Caro <david@dcaro.es>"
    a88a1edc: global: schemas and basic packaging


Bugfixes and minor changes
--------------------------
```

Again, I'd like to emphasize that release notes and changelog is not the same, the changelog is a list of changes ordered by time (a log of changes) and the release notes should be a summary of the new relevant changes added since the previous release (of course, imo).

kaplun:
@david-caro you have convinced about the Changelog Vs Release-notes :+1: 
Just: would it be possible to stick with compatible tags as defined in:
http://invenio.readthedocs.io/en/latest/technology/git.html#r2-remarks-on-commit-log-messages

In this way we can be future-compatible with Invenio.

scripts/version_manager.py
```diff
@@ -0,0 +1,464 @@
...
```
kaplun: BTW, what is the copyright of this code? Is it yours? Could you add the usual license?

david-caro: it's gplv2, same as we have here.

david-caro:
About being invenio-compatible, I'll try, though looking for a (quite common) keyword inside an internal bulletpoint restricts us to always using bulletpoints in the messages, I'll also try to map all the different keywords they have to the three semantic versioning categories somehow (feature-bug-api breaking changes).

kaplun:
:+1: Regarding bulletpoint that is what we are already using as convention in all the packages, so should be fine.

scripts/version_manager.py
kaplun: Can you indeed add the license blurb?

david-caro: It should be there..... Maybe I forgot to commit,will check tomorrow morning.

PR #2, opened 2016-09-29
api: add the validate function

david-caro:
* NEW First public utility for the module, it just validates the
  given data against the given schema (by name).

### DIFF
--- a/inspire_schemas/api.py
+++ b/inspire_schemas/api.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE-SCHEMAS.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE-SCHEMAS is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE-SCHEMAS is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE-SCHEMAS; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Public api for methods and functions to handle/verify the jsonschemas."""
+from jsonschema import validate as jsonschema_validate
+
+from .utils import LocalRefResolver, SchemaKeyNotFound, load_schema
+
+
+def validate(data, schema_name=None):
+    """Validate the given dictionary against the given schema.
+
+    :param data: Dict to validate.
+    :type data: dict
+    :param schema_name: String with the name of the schema to validate, for
+        example, 'authors' or 'jobs'. If `None` passed it will expect for the
+        data to have the schema specified in the `$ref` key.
+    :type schema_name: str
+    :return: None
+    :raises inspire_schemas.utils.SchemaNotFound: if the given schema was not
+        found.
+    :raises inspire_schemas.utils.SchemaKeyNotFound: if the given schema was
+        not found.
+    :raises jsonschema.SchemaError: if the schema is invalid
+    :raises jsonschema.ValidationError: if the data is invalid
+    """
+    if schema_name is None:
+        if '$schema' not in data:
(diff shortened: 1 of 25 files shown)


inspire_schemas/utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+    :param schema_name: String with the name of the schema to validate, for
+        example, 'authors' or 'jobs'.
+    :type schema_name: str
+    :rtype: Boolean
+    :return: The path or the given schema name.
+    :rtype: str
+    """
+    pkg_dir = os.path.dirname(__file__)
+    schema_path = os.path.join(pkg_dir, 'records', schema_name) + '.json'
```
kaplun: Here I think you should use native `pkg_resources`-based techniques. So that we could in principle ship the package as a zip one day.

inspire_schemas/api.py
```diff
@@ -0,0 +1,49 @@
...
+
+    :param data: Dict to validate.
+    :type data: dict
+    :param schema_name: String with the name of the schema to validate, for
+        example, 'authors' or 'jobs'.
+    :type schema_name: str
+    :return: None
+    :raises .utils.SchemaNotFoundException: if the given schema was not found.
+    :raises jsonschema.SchemaError: if the schema is invalid
+    :raises jsonschema.ValidationError: if the data is invalid
```
kaplun: In Invenio we normally load the schema directly from the record `data['$schema']`. Shall we default to that when `schema_name is None`?

inspire_schemas/utils.py
david-caro: I don't really see the connection though, why shouldn't it be zip-shippable already?

inspire_schemas/api.py
david-caro: makes sense, will throw an exception if `$schema` not there.

inspire_schemas/utils.py
kaplun: Because you would not find the file locally unless it would have been unzipped on the fly. Maybe you are right, but then why `pkg_resources` exist? :dancers: http://setuptools.readthedocs.io/en/latest/pkg_resources.html#resourcemanager-api

inspire_schemas/api.py
```diff
@@ -0,0 +1,58 @@
...
+        example, 'authors' or 'jobs'. If `None` passed it will expect for the
+        data to have the schema specified in the `$ref` key.
+    :type schema_name: str
+    :return: None
+    :raises inspire_schemas.utils.SchemaNotFound: if the given schema was not
+        found.
+    :raises inspire_schemas.utils.SchemaKeyNotFound: if the given schema was
+        not found.
+    :raises jsonschema:SchemaError: if the schema is invalid
+    :raises jsonschema:ValidationError: if the data is invalid
```
david-caro: hmmm. this should be a dot `jsonschema.ValidationError`

inspire_schemas/utils.py
```diff
@@ -0,0 +1,114 @@
...
+    :rtype: bool
+    :return: The path or the given schema name.
+    :rtype: str
+    """
+    schema_path = (
+        resource_filename(
+            __name__,
+            os.path.join('records', schema_name + '.json')
+        )
+    )
```
david-caro: too many extra `(`

david-caro:
@kaplun is this good on your side too?

inspire_schemas/utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+"""Public api for methods and functions to handle/verify the jsonschemas."""
+import json
+import os
+
+from jsonschema import RefResolver
+from pkg_resources import resource_filename
+
+
+class InspireSchemasException(Exception):
```
kaplun: Please, move exceptions into an `errors.py` following current conventions.

david-caro: I was taking the approach of 'only implement big-package structures when you have a big package', as well, there's only one small module that has any custom exception (not even in the api). But sure, it's quick.

david-caro:
I've added the tests here too, as well, it makes sense that a pr is finished only when the tests are in too.

david-caro:
Sorry for the noise, wanted to squash some of the commits to make the history more understandable.

david-caro:
> Additionally regarding loading sub-schema, are you sure there is not available any helper from upstream?

Not really, there's a similar implementation hidden inside jsonresolver (hidden inside a class generator function at https://github.com/inveniosoftware/jsonresolver/blob/master/jsonresolver/contrib/jsonschema.py#L38), that actually uses JSONResolver class from the core.py, that uses entry points and pluggy.PlugginManager + werkzeug.Map methods to discover and match schema paths...

Though well, I could get rid of around 10 lines of code using that approach, but probably would have to add some more lines to handle the strange api, plus it adds some new deps (worst case, all flask and related troop), a bunch of extra code (only jsonref python code is >1000 lines), so I preferred to go for the easier setup to maintain. As you pointed out the other day, we might have to depend on invenio-records in the end for some stuff, though I'd prefer avoiding it as much as possible.

PR #5, opened 2016-09-30
WIP Add tests, both unit and integration tests.

david-caro:
It also fixes a couple of issues I've found with the schemas when
autogenerating the examples for the tests.

### DIFF
--- a/inspire_schemas/records/institutions.json
+++ b/inspire_schemas/records/institutions.json
@@ -32,6 +32,9 @@
             "type": "string"
         },
         "extra_words": {
+            "items": {
+                "type": "string"
+            },
             "title": "Extra words",
             "type": "array",
             "uniqueItems": true
--- a/scripts/generate_example_records.js
+++ b/scripts/generate_example_records.js
@@ -1,0 +1,63 @@
+#!/usr/bin/env node
+//requires the json-schema-faker npm package installed
+var jsf = require('json-schema-faker')
+
+jsf.format('ISO 639-1', function(gen, schema){ return gen.randexp('^1.*$');});
+jsf.format('date', function(gen, schema){ return gen.randexp('^1.*$');});
+jsf.format('url', function(gen, schema){ return gen.randexp('^1.*$');});
+jsf.format('isbn', function(gen, schema){ return gen.randexp('^1.*$');});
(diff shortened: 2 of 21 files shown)


inspire_schemas/records/institutions.json
```diff
@@ -32,6 +32,9 @@
             "type": "string"
         },
         "extra_words": {
+            "items": {
+                "type": "string"
+            },
```
david-caro: This might not meet the reality... please advice if so.

scripts/generate_example_records.js
```diff
@@ -0,0 +1,63 @@
...
+                resolved_schema['minItems'] = 1
+            }
+            if (resolved_schema.hasOwnProperty('properties') && !resolved_schema.hasOwnProperty('anyOf')) {
+                resolved_schema['required'] = Object.keys(resolved_schema['properties'])
+            }
+            var value = unresolved_schema[key];
+            if (value !== null && typeof(value) === 'object') {
+                value = resolve_schema(unresolved_schema[key], base_path);
+                delete(value['$schema'])
+            }
```
david-caro: There's a very ugly modification of the schemas to force here to generate at least one element of each... did not find a better way to do it (that does not require rewriting/patching the generator module), any nicer alternative is welcome.

scripts/generate_example_records.js
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env node
+//requires the json-schema-faker npm package installed
+var jsf = require('json-schema-faker')
+
+jsf.format('ISO 639-1', function(gen, schema){ return gen.randexp('^1.*$');});
+jsf.format('date', function(gen, schema){ return gen.randexp('^1.*$');});
+jsf.format('url', function(gen, schema){ return gen.randexp('^1.*$');});
+jsf.format('isbn', function(gen, schema){ return gen.randexp('^1.*$');});
+jsf.format('.+, .+', function(gen, schema){ return gen.randexp('^.+, .+$');});
```
david-caro: we might want to change the formats here to something more realistic at some point, but they are not being validated on the jsonschema side either, so better fix that first.

david-caro:
I'm missing some negative tests!

david-caro:
This was moved to #2

PR #6, opened 2016-10-04
schemas: add acquisition_source to authors

jacquerie:
Let's see if I understood your `version_manager` script.

### DIFF
--- a/inspire_schemas/records/authors.json
+++ b/inspire_schemas/records/authors.json
@@ -1,6 +1,9 @@
 {
     "$schema": "http://json-schema.org/schema#",
     "properties": {
+        "acquisition_source": {
+            "$ref": "elements/acquisition_source.json"
+        },
         "advisors": {
             "items": {
                 "properties": {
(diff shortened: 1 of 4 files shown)


david-caro:
Yep, looks ok :),  I also added support for the 'NEW' tag on the commit message (invenio style), I'll open an issue to add docs about that and the tests too.

kaplun:
Acquisition source is one of those meta fields. It should be added to all schemas, until we decide how to handle meta-metadata.

jacquerie:
> Acquisition source is one of those meta fields. It should be added to all schemas, until we decide how to handle meta-metadata.

I disagree: only HEP and Author records contain this information, therefore it should only be added there. In the future **maybe** every type of record will contain it: well, that day we will update the schema. YAGNI!

kaplun:
I disagree to your disagreement. We have this information for all records albeit in the `hstRECORD` table rather than in the metadata itself. If you want, upon export, I can materialize this meta-metadata.

We keep on receiving this information and records arrive from the most disparate sources, however we fail to capture this properly on legacy but we should capture it in Labs.

That said YAGNI is really a cacophonous acronym.

Overall, indeed by the time we implement proper audit log this field will rather be deleted than added. So finally OK with your disagreement.

PR #13, opened 2016-10-05
scripts: fix generate_example_records

jacquerie:
* Makes `scripts/generate_example_records.json` work out of the box
  by adding the proper requires and making it executable.

* Also adds `node_modules` to `.gitignore` so that we can `npm install
  json-schema-faker` locally.

Signed-off-by: Jacopo Notarstefano <jacopo.notarstefano@cern.ch>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,5 @@ target/
 
 CHANGELOG
 AUTHORS
+
+node_modules
--- a/scripts/generate_example_records.js
+++ b/scripts/generate_example_records.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 //requires the json-schema-faker npm package installed
-var jsf = require('json-schema-faker')
+var fs = require('fs',
+    jsf = require('json-schema-faker'),
+    path = require('path');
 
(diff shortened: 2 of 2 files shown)


jacquerie:
@david-caro, thoughts?

kaplun:
Oh! Why was this implemented in JS? (sorry not really related with this library but I am noticing it just now). There must be an equivalent library in Python.

kaplun:
E.g. https://pypi.python.org/pypi/fake-factory

jacquerie:
> E.g. https://pypi.python.org/pypi/fake-factory

This is missing the JSON Schema part. Here's a relevant issue: https://github.com/Julian/jsonschema/issues/259

Maybe add an issue about this, just to add it to the record.

david-caro:
this looks ok for me, I did it in json because I did not find a nice library for python that did it, even with the json one, I was unable to really get the full examples from the default api... this small js script took me half an evening (did not want to spend much time as this was blocking a bunch of other things).

Open to any suggestion :)

kaplun:
Looks like there's no python JSONSchema object generator. So indeed what you did was the most practical thing :+1:

PR #16, opened 2016-10-05
WIP schemas: add missing types

jacquerie:
@harunurhan reported that we were missing some types in the `ids` element.

We're probably missing more, so, since you can make your editor crash on missing types (by removing the `undefined` check), why don't you try it on the examples in `tests/integration/fixtures` and report back all the other misses?

### DIFF
--- a/inspire_schemas/records/elements/id.json
+++ b/inspire_schemas/records/elements/id.json
@@ -103,7 +103,8 @@
                     "pattern": "CERN-\\d+",
                     "type": "string"
                 }
-            }
+            },
+            "type": "object"
         },
         {
             "properties": {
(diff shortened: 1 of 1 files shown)


kaplun:
I think `domapping` tool also, as a side effect, spits out missing types...

harunurhan:
@jacquerie it already crashes in those cases, I will try those fixture with their schema on editor see if it works!

david-caro:
That is strange, shouldn't this make the tests fail too? Or the types might be inferred by default?

david-caro:
If that's the case and we need the types, we might want to add a small test that does the check.

jacquerie:
> That is strange, shouldn't this make the tests fail too? Or the types might be inferred by default?

No, a JSON Schema does not need a `type`. For instance, `{}` is a valid JSON Schema that accepts any well-formed JSON document. For `record-editor`, we will need `type`s everywhere, so yeah, we should be testing for that. Actually, let me write that...

PR #17, opened 2016-10-06
WIP api: add resolve

jmartinm:
* Adds resolve function that given a schema name returns the schema
  with all $ref resolved.

Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>


jmartinm:
This is not needed since the functionality was implemented already in `inspire-schemas`

PR #23, opened 2016-10-10
hep_schema: known categories are now enums

david-caro:
* ARXIV and INSPIRE scheme categories have now enums with valid
  options.
* Other scheme types (APS and POS) can have any values, but no other
  schemes are allowed.


david-caro:
Waiting to resolve current logging issues on record migration on nightly so we will have insight on what this breaks on the current records there.

david-caro:
Actually, we can merge this, as we are using semantic versioning, this will not get deployed on nightly until we change the requirements on inspire-next :)

inspire_schemas/records/elements/field.json
```diff
@@ -1,42 +1,98 @@
...
+        "sources": {
             "enum": [
                 "INSPIRE",
                 "submitter",
                 "conference",
                 "curator",
                 "publisher",
+                "arxiv",
+                "pos",
+                "asp",
```
jacquerie: What are `pos` and `asp`?

inspire_schemas/records/elements/field.json
```diff
@@ -1,42 +1,98 @@
...
-            "type": "string"
         }
     },
-    "type": "object"
+    "anyOf": [
+        {
+            "properties": {
+                "scheme": {
+                    "type": "string",
+                    "enum": ["INSPIRE"]
```
jacquerie: Mh, maybe we should add something like `prettifyjson` here.

tests/integration/fixtures/authors_example.json
```diff
@@ -1,182 +1,172 @@
 {
```
jacquerie: Maybe I misunderstood how you meant to use your `generate_records` script, but here IMO it would be better to only have the changes about the thing that you really changed, rather than a completely new record...

inspire_schemas/records/elements/field.json
david-caro: PoS = Proceedings of science
ASP is a typo, it should be aps -> American Physical Society

Essentially, two spiders that we already have that use this.

inspire_schemas/records/elements/field.json
david-caro: I don't care much, what would it improve?

tests/integration/fixtures/authors_example.json
david-caro: Well, these examples are meant to be to ensure backwards compatibility, so you should not use them to see what has changed. To see the changes you should look to the schema itself.

I agree though that it would be a plus to have the scripts regenerate only the portions that changed, but I see no easy way to do it (that does not involve some non-trivial programming on the ugly js script).

For now I just regenerate the whole lot each time to make sure it's maintained complete.

inspire_schemas/records/elements/field.json
david-caro: Found an easy way to pretiffy all json, sent a new patch with a script for it + pretified everything with it :)

david-caro: Hmmm... I'll add a test for it too so we don't forget pretifying at the end, maybe even a hook-like script to add to the git before commit or something might work too...

david-caro:
Let's merge this, it will not break anything on inspire-next, but allows testing the related changes on hepcrawl.

PR #30, opened 2016-10-27
hep: Insertion of fields `$schema` and `control_number` in hep.json.

zzacharo:
- Insertion of the `$schema` and `control_number` fields in the hep schema
  as it is required for using invenio-records-rest PUT request upon updating
  a single record.

Signed-off-by: Zacharias Zacharodimos zacharias.zacharodimos@cern.ch

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -2,6 +2,10 @@
     "$schema": "http://json-schema.org/schema#",
     "description": "An article or thesis or book or...",
     "properties": {
+        "$schema": {
+            "format": "url",
+            "type": "string"
+        },
         "abstracts": {
             "items": {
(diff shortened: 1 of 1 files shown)


inspire_schemas/records/hep.json
```diff
@@ -254,6 +258,9 @@
             "type": "array",
             "uniqueItems": true
         },
+        "control_number": {
+            "type": "string"
```
jmartinm: I think it should be `number`. Zenodo also uses a field to store this old Invenio identifiers and defined them as numbers - https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json

jmartinm:
Apart from the comment, LGTM.

PR #32, opened 2016-10-31
authors: reintroduce old_emails

jmartinm:
* Since we still need to import/export from legacy, reintroduce
  explicit old_emails and emails in positions to avoid losing information.


jacquerie:
...this broke the build again, because https://github.com/inspirehep/inspire-schemas/pull/35 wasn't merged before this : (

PR #34, opened 2016-11-03
hep: Insertion of fields `$schema` and `control_number` in hep.json.

zzacharo:
* Insertion of the `$schema` and `control_number` fields in the hep schema
   as it is required for using invenio-records-rest PUT request upon updating
   a single record.

Signed-off-by: Zacharias Zacharodimos <zacharias.zacharodimos@cern.ch>

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -2,6 +2,10 @@
     "$schema": "http://json-schema.org/schema#",
     "description": "An article or thesis or book or...",
     "properties": {
+        "$schema": {
+            "format": "url",
+            "type": "string"
+        },
         "abstracts": {
             "items": {
                 "properties": {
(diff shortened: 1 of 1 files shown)


jacquerie:
This was a breaking change (https://travis-ci.org/inspirehep/inspire-next/builds/173215901), and should have had something in the commit message to indicate that. This was the main reason to separate `inspire-schemas` in its own package!

jmartinm:
Spotting that it is a breaking change is not obvious though. In principle adding fields to the schema should not be a breaking change as far as I know.

Not sure if there is a way to spot it automatically.

jacquerie:
Adding any constraint on a field that someone is already populating is _always_ a breaking change, because you are restricting the possible values it can have, therefore potentially producing new invalid records.

If instead you remove constraints (e.g. by adding things to an `enum`) then the number of invalid records can only decrease, so it's not a breaking change.

PR #44, opened 2016-11-25
Refactor the field_categories

rikirenz:
This change is:

- removing the field: `field_categories`
- adding 2 field: `external_field_categories` and `inspire_field_categories`

In order to maintain different lists for the categories of INSPIRE and the categories that we receive from external sites, during the articles collection

### DIFF
--- a/scripts/generate_example_records.js
+++ b/scripts/generate_example_records.js
@@ -10,18 +10,27 @@ jsf.format('url', function(gen, schema){ return gen.randexp('^1.*$');});
 jsf.format('isbn', function(gen, schema){ return gen.randexp('^1.*$');});
 jsf.format('.+, .+', function(gen, schema){ return gen.randexp('^.+, .+$');});
 
-function resolve_schema(unresolved_schema, base_path) {
+
+function extract_definition(definitions_object, defs_path) {
+  var keys_array = defs_path.split('/');
+  return definitions_object[keys_array[1]];
+}
+
+
+function resolve_schema(unresolved_schema, base_path, definitions) {
+    definitions = unresolved_schema['definitions'] || definitions;
+
     var keys = [];
     var resolved_schema = {};
     if (Array.isArray(unresolved_schema)) {
-        new_array = []
-            for (var value of unresolved_schema) {
-                if (value !== null && typeof(value) === 'object') {
-                    new_array = new_array.concat(resolve_schema(value, base_path))
-                } else {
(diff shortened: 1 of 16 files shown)


rikirenz:
Ready for review! @david-caro

scripts/generate_example_records.js
```diff
@@ -10,28 +10,39 @@ jsf.format('url', function(gen, schema){ return gen.randexp('^1.*$');});
...
-function resolve_schema(unresolved_schema, base_path) {
+
+function set_definitions(current_definitions, parent_definitions){
+    if (current_definitions == null) return parent_definitions;
+    else return current_definitions;
+}
+
+
+function resolve_schema(unresolved_schema, base_path, definitions) {
+    definitions = set_definitions(unresolved_schema['definitions'], definitions);
```
david-caro: I would use select/pick instead of set for the name of the method, as usually the methods that have 'set' on the name are expected to modify on of the parameters or some global, setting a property on them.

scripts/generate_example_records.js
```diff
@@ -10,28 +10,39 @@ jsf.format('url', function(gen, schema){ return gen.randexp('^1.*$');});
...
         if (key === '$ref' && typeof(unresolved_schema[key]) === 'string' && unresolved_schema[key][0] !== '#') {
             var schema_path = base_path + "/" + unresolved_schema[key];
             var new_base_path = path.dirname(schema_path);
             var element_schema = JSON.parse(fs.readFileSync(schema_path));
-            var resolved_element_schema = resolve_schema(element_schema, new_base_path);
+            var resolved_element_schema = resolve_schema(element_schema, new_base_path, definitions);
             delete(resolved_element_schema['$schema'])
             return resolved_element_schema
+        } else if (key === '$ref' && typeof(unresolved_schema[key]) === 'string' && unresolved_schema[key][0] === '#') {
+            return definitions
```
david-caro: shouldn't this return `resolve_schema(extract_definition(definitions, defs_path), schema_path, definitions)`? Where the defs_path is the path part of the '#key1/key2' and the `extract_definiton` just extracts the value for the key it references?

rikirenz:
I think you can merge!

david-caro:
you should rebase on latest master too, to get rid of that extra commit that's shown.

rikirenz:
@david-caro rebased and tests have passed.

Issue #41, opened 2016-11-18
Privilege INSPIRE categories


kaplun:
Currently INSPIRE categories are stored at the same level of arXiv categories and other externally provided categories.

However, while cataloger are not supposed to touch externally provided categories, they are instead expected to curate INSPIRE categories. This poses INSPIRE categories in a privileged place.

It is proposed to move INSPIRE categories on a dedicated field, so that they are easier to edit (e.g. autocomplete can be enforced on the exact INSPIRE categories).

rikirenz:
cc: @david-caro

kaplun:
So we could split it in `INSPIRE_categories` and `external_categories`. INSPIRE_catagories will not need any scheme or source. So just list of strings.

david-caro:
I'm always confused about caps and lowercase, isn't it better to keep everything in lowercase? (As if they were variable names in python)

kaplun:
Sure!

kaplun:
@inspirehep/inspire-content @inspirehep/inspire-dir 

Given the following facts:
* In the last 6 months only ~0.01% users queried INSPIRE using arXiv categories (or any category at all)
* arXiv categories are not available for all types of documents

Given the following thoughts:
* Faceting for ~10 categories (INSPIRE) is nicer than faceting for ~30 (arXiv)
* Using neural network (magpie) to guess ~10 categories (INSPIRE) is more precise and reliable than ~30 categories (arXiv)

It is proposed here to simplify the data model and hence curation inteface in the following way:
* [ ] store only INSPIRE categories
* [X] upon ingestion external categories (e.g. arXiv) are mapped to INSPIRE ones
* [X] if no external category we use magpie to guess one based on abstract
* [ ] upon migration from legacy we preserve INSPIRE categories and we map existing arXiv ones into INSPIRE.
* [ ] Present to users a facet based on INSPIRE categories.

david-caro:
So we only use magpie if there are no external categories?

kaplun:
If we don't have any INSPIRE category (they might come through a different mapping in hepcrawl, for example)

salmele:
I *strongly* disagree.

Our user community understands what hep-th means, for instance, and ought to be able to search accordingly

Possibly, we've a 'topic' search, which has all of INSPIRE present 'categories'. And arXiv-category search, which is meant to work only on arXiv content, and not second-guessed INSPIRE attribution

jacquerie:
[...]

PR #47, opened 2016-12-05
Refactoring categories

rikirenz:
We are going to create 2 fields:

- One for the INSPIRE categories
- One for the arXiv categories


rikirenz:
Ready for review! @david-caro

inspire_schemas/records/elements/arxiv_field.json
```diff
@@ -0,0 +1,8 @@
+{
```
david-caro:
```
{
    'items': {
        'oneOf': [
            {
                'enum': ['one', 'two'],
                'type': 'string'
            }
        ]
    },
    'type': 'array',
    'uniqueItems': True
}

```

rikirenz:
let's try again!

david-caro:
This looks great now 👍 , but we should wait for #41 to see if this is actually what it's needed before merging, though we can start modifying the rest of the code (inspire-next + hepcrawl) to adapt to this in the meantime, as if not like this, it will be quite similar I think.

david-caro:
I forgot! this is a non-backwards compatible change, so you should use this:
http://autosemver.readthedocs.io/en/latest/usage.html#major-change

rikirenz:
@david-caro again new changes

david-caro:
+ 💯 !

PR #48, opened 2016-12-12
schemas: add ids to institutions

jacquerie:
As mentioned in https://github.com/inspirehep/inspire-next/pull/1807.

### DIFF
--- a/authors_example.json
+++ b/authors_example.json
@@ -1,0 +1,275 @@
+{
+    "acquisition_source": {
+        "date": "sunt deserunt E",
+        "email": "exercitation",
+        "method": "sunt dolore commodo dolor",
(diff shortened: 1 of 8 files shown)


david-caro:
The error on travis looks unrelated... on the other side, I think you should regenerate the example records so they include the new field.

jacquerie:
> The error on travis looks unrelated...

Looks like a regression in `docutils`. We could pin 0.12, I guess?

> I think you should regenerate the example records so they include the new field

Sure! Done.

jacquerie:
Or we could wait for Sphinx 1.5.1: https://github.com/sphinx-doc/sphinx/issues/3229

PR #55, opened 2017-01-13
authors: add missing schema fields

### LINKED ISSUE #54: author schemas missing fields
Following fields are being used in author forms but are not on the schema. One of the side effects is that those fields are not visible in the record editor, so not editable from the Holding Pen.

* [ ] _private_note
* [ ] collections
* [ ] _degree_type
* [ ] _rank

jmartinm:
* (closes #54).

Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>

### DIFF
--- a/inspire_schemas/records/authors.json
+++ b/inspire_schemas/records/authors.json
@@ -1,12 +1,18 @@
 {
     "$schema": "http://json-schema.org/schema#",
     "properties": {
+        "_private_note": {
+            "type": "string"
+        },
         "acquisition_source": {
             "$ref": "elements/acquisition_source.json"
         },
         "advisors": {
             "items": {
                 "properties": {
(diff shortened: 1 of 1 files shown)


inspire_schemas/records/authors.json
```diff
@@ -40,6 +46,20 @@
...
+            "items": {
+                "properties": {
+                    "primary": {
+                        "type": "string"
+                    }
+                },
+                "type": "object"
+            },
+            "minItems": 1,
+            "title": "Collection",
```
jacquerie: IMO we don't `title`s and `description`s that repeat the key name.

jmartinm: I copy pasted from `hep.json`. Agree that it does not bring any useful info.

PR #56, opened 2017-01-16
hep: fix arXiv eprint regexp

kaplun:
* FIX Fixes arXiv eprint regexp, so that it correctly catches
  optional subcategories.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -52,6 +52,7 @@
             "$ref": "elements/acquisition_source.json"
         },
         "arxiv_eprints": {
+            "description": "Read only meta-data (identifiers and categories) coming from arXiv",
             "items": {
                 "properties": {
                     "categories": {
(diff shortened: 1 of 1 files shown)


kaplun:
@inspirehep/inspire-content we are here deciding that arXiv eprints stored in literature will be stored without the arXiv prefix (that can be added upon export), and without versioning, since, for INSPIRE several versions are going to be represented in one entry.

inspire_schemas/records/hep.json
```diff
@@ -61,10 +62,14 @@
                         "type": "array"
                     },
                     "value": {
-                        "pattern": "\\d{4}.\\d{4,5}|\\w+-\\w+/\\d+|\\w+/\\d+",
+                        "description": "arXiv eprint number, e.g. math.DG/0307245 or 1701.01431",
+                        "pattern": "\\d{4}.\\d{4,5}|[\\w.]+(-[\\w.]+)?/\\d+",
```
david-caro: something like `123..../321` is valid?

david-caro: (there's no need to get a hugely complicated 100% valid regexp right now, but if this is known not to be 100% compliant, we should add an issue at least)

kaplun: It's just that the previous regex would have rejected valid ones such as `math.DG/03343`. I would say we almost do not need a regexp, since this field should be read-only (since it's pure info coming from arXiv and shouldn't be touched by catalogers). However, it's good to have a regexp in order to ensure that all workflows respect the same convention (e.g. no leading `arXiv:` prefix or `v1` suffix).

PR #57, opened 2017-01-16
hep: finalize author structure

kaplun:
* INCOMPATIBLE Restructures the way to store CRediT roles and INSPIRE
  roles.

* NEW Introduces support for raw values.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -71,25 +71,22 @@
             "uniqueItems": true
         },
         "authors": {
-            "description": "List with all the authors",
+            "description": "Anybody who has acted on a given document (including supervisors, editors, etc.)",
             "items": {
                 "properties": {
                     "affiliations": {
                         "items": {
                             "properties": {
                                 "curated_relation": {
-                                    "description": "Did a cataloguer proof-checked the recid?",
-                                    "title": "The affiliation is curated?",
+                                    "description": "Does this affiliation point to the correct Institution?",
                                     "type": "boolean"
                                 },
                                 "record": {
                                     "$ref": "elements/json_reference.json",
-                                    "description": "URI for the Institution collection record",
-                                    "title": "URI for the Institution collection record"
+                                    "description": "URI for the Institution collection record"
                                 },
                                 "value": {
-                                    "description": "The affiliation as it appears on the paper",
-                                    "title": "Name of institution",
+                                    "description": "The normalized affiliation, i.e. the institution identifier.",
                                     "type": "string"
                                 }
                             },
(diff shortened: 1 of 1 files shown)


kaplun:
@inspirehep/inspire-content this the proposal on how to represent signatures in papers.

The idea is to put in the same field supervisors and editors, but tag them properly, and additionally support CRediT roles for authors.

inspire_schemas/records/hep.json
```diff
@@ -133,16 +140,28 @@
...
                             "type": "object"
                         },
                         "type": "array",
                         "uniqueItems": true
                     },
                     "curated_relation": {
                         "default": false,
-                        "description": "Was this signature actually claimed or proof-checked by cataloguer?",
+                        "description": "Does this signature point to the correct author profile?",
                         "title": "The relation is curated?",
```
jacquerie: Should we deduce that, if it's `false`, then we know that it's incorrect? Then why are we storing it?

IMO these `description`s and `title`s are just noise.

jacquerie:
This reverts a previous decision to use only CRediT roles, reverting progress in Inspire.

michamos:
> This reverts a previous decision to use only CRediT roles, reverting progress in Inspire.

We decided on Inspire roles after careful review of CRediT roles with @kaplun and @annetteholtkamp :

## Editor

This is for editors of conference proceedings & special issues of journals.
The closest CRediT would be "writing - review & editing", but the [definition](http://dictionary.casrai.org/Contributor_Roles/Writing_%E2%80%93_review_%26_editing) is (emphasis mine):

> Preparation, creation and/or presentation of the published work **by those from the original research group**, specifically critical review, commentary or revision – including pre- or post-publication stages.

An editor is usually not (or only accidentally) from the original research group of the paper.

## Supervisor

We want to be able to distinguish a thesis supervisor, who is **not** author of the thesis, from the CRediT role, which would mean authorship.

## Author

All other cases.

kaplun:
This comes from the fact that CRediT has a meaning of editor and supervisor different than INSPIRE ones. And @michamos and @annetteholtkamp both remarked the difficulties that would arise should we use just credit roles when e.g. describing author of a collaboration. I let them better explain it.

inspire_schemas/records/hep.json
kaplun: If if's false: it's not "curated". 

RE: description: proposal are accepted on how to make them not noise.
@michamos suggests these should be there to guide API users on the meaning of each fields.
Alternatively these could be used in the editor to guide the cataloger on how to fill them.

I am not sure about the best solution.
But sometimes they are really needed in order to make explicit certain inner thought.

jacquerie: > proposal are accepted on how to make them not noise.

Remove them. If I have a boolean called `curated_relation` then I can deduce that `curated_relation = False` means that this relation was not curated. I don't need a `title` that duplicates this obvious fact. I need even less a `description` that makes me wonder "uh, if this is false maybe this means that this is incorrect...".

jacquerie:
> We decided on Inspire roles after careful review of CRediT roles

Fair enough.

inspire_schemas/records/hep.json
michamos: problem is `curated_relation` doesn't always mean the same thing dependent on context: it is an indication that something has been checked by someone, but the details seem to vary.

kaplun: Sometimes there are nested relations (e.g. affiliation within author). Then there is a curated_relation for the author, and a nested curated_affiliation for each attached affiliation. Could be confusing about which is which (I know it's nested, but this might mix up in the eye of the user).

Anyway I don't have a strong opinion.

jacquerie: > problem is `curated_relation` doesn't always mean the same thing dependent on context: it is an indication that something has been checked by someone, but the details seem to vary.

(writing down the result of the conversation that we had)

Sometimes `curated_relation` means "I proof checked that this is pointing to the right record", sometimes it means "I amended the output of that script so that it is correct, I'm instructing the script to not modify it again".

inspire_schemas/records/hep.json
```diff
@@ -100,22 +97,32 @@
                         "type": "array",
                         "uniqueItems": true
                     },
+                    "raw_affiliations": {
+                        "items": {
```
david-caro: This items should not be here, or it should have the `"type": "array"` key at the same level.
Do we have more than one source of affiliations?

kaplun: Unfortunately, as usual in case of merge: if we have a raw affiliation from arXiv, and one from the publisher.

kaplun: Indeed array is missing!

kaplun: BTW: Do we have a test, ensuring that the schemas are jsonschema compliant?

michamos:
This is blocking #58

inspire_schemas/records/hep.json
jacquerie: This test is still in Inspire: https://github.com/inspirehep/inspire-next/blob/a6c641e860a9e7c357e30af5733edef9546afb76/tests/unit/records/test_records_jsonschemas.py

kaplun: Thanks. Ticketized in #62

PR #68, opened 2017-01-20
Many fixes

kaplun:


inspire_schemas/records/authors.json
```diff
@@ -17,13 +17,7 @@
...
                     },
                     "degree_type": {
-                        "enum": [
-                            "PhD",
-                            "Master",
-                            "Bachelor",
-                            "Other"
-                        ],
-                        "type": "string"
+                        "$ref": "elements/json_reference.json"
```
michamos:
```
"$ref": "elements/degree_type.json"
```

inspire_schemas/records/elements/cnum.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "pattern": "C\\d\\d-\\d\\d(-\\d\\d)?(\\.\\d+)?",
```
michamos: Do we have CNUMs without starting day? I couldn't find any.

inspire_schemas/records/hep.json
```diff
@@ -659,14 +651,17 @@
                         "description": "First page",
                         "type": "string"
                     },
+                    "parent_isbn": {
+                        "type": "string"
```
michamos: Shouldn't the ISBN be validated with a regex?

inspire_schemas/records/hep.json
```diff
@@ -668,6 +662,17 @@
...
         },
+        "publication_type": {
+            "enum": [
+                "introductory",
+                "lecture",
+                "review"
+            ],
+            "type": "string"
+        },
+        "published": {
```
michamos: `refereed`?

inspire_schemas/records/elements/source.json
```diff
@@ -0,0 +1,4 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "type": "string"
```
jmartinm: Is it worth separating into its own file just to put `type: string` ? :P it does not seem to be its own element.

inspire_schemas/records/hep.json
kaplun: This was really do distinguish between preprint and published.
`refereed` would be a different flag. See: https://github.com/inspirehep/inspire-schemas/issues/31#issuecomment-274445133

inspire_schemas/records/elements/source.json
kaplun: It is because I would like to make it an `enum`. But I don't have at the moment the energy to dig legacy for the allowed values.

Ideally these would be: `CDS`, `arXiv`, `Elsevier`, `Springer`, `User`, `APS`...

PR #69, opened 2017-01-24
schemas: add noAdditionalProperties everywhere

### LINKED ISSUE #26: schemas: add noAdditionalProperties everywhere
Continues https://github.com/inspirehep/inspire-next/pull/1504.

kaplun:
* BETTER Adds noAdditionalProperties everywhere.
  (closes #26)

### DIFF
--- a/inspire_schemas/records/authors.json
+++ b/inspire_schemas/records/authors.json
@@ -1,5 +1,6 @@
 {
     "$schema": "http://json-schema.org/schema#",
+    "noAdditionalProperties": true,
     "properties": {
         "_private_note": {
             "type": "string"
@@ -9,6 +10,7 @@
         },
         "advisors": {
             "items": {
+                "noAdditionalProperties": true,
                 "properties": {
                     "_degree_type": {
                         "type": "string"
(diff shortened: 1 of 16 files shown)


inspire_schemas/records/authors.json
```diff
@@ -264,11 +271,12 @@
...
-                "date_verified": {
-                    "format": "date",
-                    "type": "string"
-                },
+                "noAdditionalProperties": true,
                 "properties": {
+                    "date_verified": {
+                        "format": "date",
+                        "type": "string"
+                    },
```
kaplun: This I realized was a misplaced property.

jacquerie:
This needs a big fat incompatible tag...

kaplun:
Does it? I think it's compatible. Rules that have generated non-defined properties where erring on the wrong side anyway.

But sure, I can add the incompatible tag.

jacquerie:
Mostly for dividing development in two: first we generate what's in the
schema at version N, then in version N+1 we avoid what's not in the schema.

Otherwise you have to do both at the same time...

PR #70, opened 2017-01-25
utils.get_schema_path: add tests and refactor

david-caro:

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -26,14 +26,15 @@
 import json
 import os
 
-from urlparse import urlsplit
-
 from jsonschema import RefResolver
 from pkg_resources import resource_filename
+from six.moves.urllib.parse import urlsplit
 
 from .errors import SchemaNotFound
 
+
 _schema_root_path = os.path.abspath(resource_filename(__name__, 'records'))
+
 
 class LocalRefResolver(RefResolver):
     """Simple resolver to handle non-uri relative paths."""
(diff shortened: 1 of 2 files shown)


david-caro:
As penitence for pushing the previous commit directly to master, I just refactored a bit and added some better tests for it.

david-caro:
@kaplun without this the release is not really made (it did not pass the tests)

@jacquerie this makes up a bit for the previous push :)

kaplun:
Did @jacquerie noticed!?

inspire_schemas/utils.py
```diff
@@ -53,26 +54,44 @@ def get_schema_path(schema):
...
     :return: The path or the given schema name.
     :rtype: str
     """
-    path = original_path = os.path.normpath(urlsplit(schema).path)
-    if path.startswith(os.path.sep):
-        path = path[1:]
-    if not path.endswith('.json'):
-        path += '.json'
+    def _strip_first_path_elem(path):
+        """Pass doctests.
```
kaplun: Pass doctest!?

david-caro: Yep, we have it enabked and it forces you to write a 'less than one line' docstring that ends with a dot. But I feel really stupid when the first line of the docstring is almost the same as the function name...

Like:
```
def do_something():
    """ Does something.

    Something means blabla...
    """
```

I'll change it if it hurts anyone's eyes :/

kaplun: Formally would be better `Does something`, but I believe a good laugh like the above would allow for a more proficient development day.

kaplun:
LGTM

PR #71, opened 2017-01-26
hep: minor enum fixes

kaplun:

### DIFF
--- a/inspire_schemas/records/elements/degree_type.json
+++ b/inspire_schemas/records/elements/degree_type.json
@@ -1,13 +1,13 @@
 {
     "$schema": "http://json-schema.org/schema#",
     "enum": [
-        "Other",
-        "Diploma",
-        "Bachelor",
-        "Laurea",
-        "Master",
-        "PhD",
-        "Habilitation"
+        "other",
+        "diploma",
+        "bachelor",
+        "laurea",
+        "master",
+        "phd",
+        "habilitation"
     ],
     "title": "Degree type",
     "type": "string"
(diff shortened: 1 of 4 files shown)


inspire_schemas/records/elements/degree_type.json
```diff
@@ -1,13 +1,13 @@
...
-        "Master",
-        "PhD",
-        "Habilitation"
+        "other",
+        "diploma",
+        "bachelor",
+        "laurea",
+        "master",
+        "phd",
+        "habilitation"
```
kaplun: Note: I am forcing this to be lower case because it's easier to the normalize when coming from MARCXML (and Inenio 1 is case insensitive)

david-caro: But then it will show up in lowercase (or best in normal case) on the ui side no? Because if we would want to make it to show correctly we will need some extra mapping (actualli, only phd is special).

kaplun: I think this is a general good habit that we should have, i.e. not to assume that a given value in an enum has to be displayed as-is in front of the user.

jacquerie: > I think this is a general good habit that we should have, i.e. not to assume that a given value in an enum has to be displayed as-is in front of the user.

I agree 100%: let's not mix view concerns in our data model, because the domain we're modeling is complex enough as it is : )

david-caro: Hmm... what I don't like is that while this makes it simpler on the model level, overall it increases the complexity by forcing us to add a transformation layer on the ui side (that is, keeping yet another model structure for ui plus the transformations in between them).

PR #74, opened 2017-01-31
Refactor things

michamos:


inspire_schemas/records/elements/sourced_value.json
```diff
@@ -0,0 +1,15 @@
+{
```
michamos: this is a `sourced_value` item instead of `sourced_values` array so that we can have different descriptions for the different fields. This is an unfortunate limitation of JSON schema.

inspire_schemas/records/elements/sourced_value.json
```diff
@@ -0,0 +1,15 @@
...
+        },
+        "value": {
+            "type": "string"
+        }
+    },
+    "required": [
+        "value"
+    ],
+    "type": "object"
+}
```
kaplun: This is a good idea :tm:
Can you do the same substitution in the other schemas, if possible?

PR #75, opened 2017-02-01
schemas: kill collections

michamos:
* INCOMPATIBLE kills collections for hep and authors.

    - In hep, collections have been superseded by document_type,
    publication_type, citeable, refereed, special_collections.

    -For authors, it is completely useless as the existing values on
    legacy are 'USEFUL' and 'HEPNAMES', both of which have special
    meaning.

* INCOMPATIBLE also pluralizes special_collections, which is an array.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -240,21 +240,6 @@
             "type": "array",
             "uniqueItems": true
         },
-        "collections": {
-            "items": {
-                "additionalProperties": false,
-                "properties": {
-                    "primary": {
-                        "type": "string"
-                    }
-                },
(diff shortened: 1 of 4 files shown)


inspire_schemas/records/hep.json
```diff
@@ -713,7 +698,7 @@
             "description": "Url of the record itself",
             "title": "Url of the record"
         },
-        "special_collection": {
```
kaplun: I can't believe I wrote it singular :hankey:

kaplun:
@jacquerie although this is an INCOMPATIBLE change, it's a minor one from the implementation POW, and allows for even more consistency.

PR #77, opened 2017-02-03
hep: require at least one arXiv category

michamos:
* INCOMPATIBLE adds "categories" to the required keys of "arxiv_eprints"
and force it to have at least one element.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -61,7 +61,9 @@
                         "items": {
                             "$ref": "elements/arxiv_categories.json"
                         },
-                        "type": "array"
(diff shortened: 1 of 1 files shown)


kaplun:
So this is formally INCOMPATIBLE, but from the code point of view is not adding/amending any field. So no  change is implied in the inspire-next.

On the other hand, when migrating from legacy, all those records that do not have categories, for whatever reason, will fail.

It's thus necessary to spot all of them and correct them (either manually or by a script).

michamos:
> On the other hand, when migrating from legacy, all those records that do not have categories, for whatever reason, will fail.

:+1: Only 4 records, doing that right now

PR #72, opened 2017-01-27
hep: improved enums

kaplun:
* INCOMPATIBLE Replaces isbns.material with isbns.medium.

* NEW Introduces an enum for acquisition_source methods.

Signed-off-by: Samuele Kaplun <samuele.kaplun@cern.ch>

PR #78, opened 2017-02-07
Schema 15

michamos:


kaplun:
@michamos can you also take and superseed my #72? So that we can then change in one go all the things that need to be changed?

michamos:
@spirosdelviniotis some of the changes here are what we were discussing

PR #81, opened 2017-02-10
lit_builder: add class to handle hep records

rikirenz:
We are adding a literature_builder in order to create `hep records`. The builder is not totally complete.  
we are missing:

- Some fields and properties

- Unit tests


inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+from inspire_schemas.api import validate
+
+""" Literature Suggestion builder"""
+
+
+class LiteratureBuilder(object):
+
+    def __init__(self, source):
+        self.obj = {}
+        self.source = source
```
jmartinm: What is this source? And why it needs to be set when creating the class?

david-caro: 👍 for adding to the docs.

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+        :type source: string
+        """
+        self._ensure_field('inspire_categories', [])
+
+        self.obj['inspire_categories'].extend([{
+            'term': category,
+            'source': self._get_source(source),
+        } for category in subject_term])
+
+    def add_private_notes(self, private_notes=None,
```
jmartinm: I am not sure about this, it seems to be quite particular to the literature submission form and in the end all the parameters end up in `private_notes`.

I think it would be better to just receive a list of `private_notes` you want to add. The caller of this method will have to clean up their mess and extract the private notes first

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+
+        has_page_or_artid = any([
+            key in self.obj['publication_info'] for key in (
+                'page_start', 'page_end', 'artid')])
+
+        if has_pub_info and has_page_or_artid:
+            self.obj['citeable'] = True
+            self.obj['refereed'] = True
+
+    def add_preprint_created(self, preprint_created):
```
jmartinm: `add_preprint_created_date`

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+        """
+        :param preprint_created: preprint date.
+        :type preprint_created: string. A date format is required (yyyy-mm-dd).
+        """
+
+        self._ensure_field('imprints', [])
+
+        self.obj['imprints'].append({
+            'date': preprint_created
```
jmartinm: I guess it is more a data model question but this field called `date` seems a bit ambiguous. Is it creation date? Upload date?

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+        self._ensure_field('accelerator_experiments', [])
+
+        self.obj['accelerator_experiments'].append({
+            'experiment': experiment
+        })
+
+    def add_language(self, language):
+        """
+        :param languge: language for the current artifact.
+        :type languge: string (2 characters)
```
jmartinm: you can do as a first line:

```python
assert len(language) == 2
```

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+    def add_license_url(self, license_url):
+        """
+        :param license_url: url for the description of the license.
+        :type license_url: string
+        """
+        self._ensure_field('license', [])
+
+        self.obj['license'].append({'url': license_url})
+
+    def add_public_note(self, public_notes, source=None):
```
jmartinm: you seem to pass only one? so `public_note`

inspire_schemas/literature_builder.py
david-caro: I'm not sure that should be done here... it's the schema the one that should be checking things

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+        """
+        self._ensure_field('arxiv_eprints', [])
+
+        arxiv_eprints = {}
+        arxiv_eprints['value'] = arxiv_id
+
+        # arXiv categories
+        arxiv_eprints['categories'] = arxiv_categories.split()
+        if len(arxiv_id.split('/')) == 2:
+            arxiv_eprints['categories'].append(arxiv_id.split('/')[0])
```
kaplun: Be careful: it is very likely that this category is already in the list. So you should first check this, before creating duplicates.

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+        :param doi: doi for the current document.
+        :type doi: string
+
+        :param source: source for the current doi.
+        :type source: string
+        """
+        self._ensure_field('dois', [])
+
+        self.obj['dois'].append({
+            'value': doi,
```
kaplun: I wonder: should we do some normalization here? Or assume the client code will do it?
Pros: client code will be simpler and can delegate it: i.e. should not care about doi:10... dx.doi.org/10. http://dx.doi.org/...
Cons: client risk to become more stupid. Or we risk to normalize twice for nothing.

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+            'source': self._get_source(source),
+        })
+
+    def add_authors(self, authors, supervisors=None):
+        """
+        :param authors: authors for a given artifact.
```
kaplun: What is the expected structure of an `author`? Shall we have add a classmethod to generate an author from its components so that it can then be then passed here?

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+        publication_item = {}
+        for key in ('cnum', 'artid', 'page_end', 'page_start',
+                    'journal_issue', 'journal_title',
+                    'journal_volume', 'year'):
+            if locals()[key] is not None:
+                publication_item[key] = locals()[key]
+
+        if page_start and page_end:
+            try:
+                publication_item['number_of_pages'] = int(page_end) - int(page_start) + 1
```
kaplun: Be careful: this has to be done only if `number_of_pages` wasn't already set.

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+            key in self.obj['publication_info'] for key in (
+                'year', 'journal_issue', 'journal_volume')])
+
+        has_page_or_artid = any([
+            key in self.obj['publication_info'] for key in (
+                'page_start', 'page_end', 'artid')])
+
+        if has_pub_info and has_page_or_artid:
+            self.obj['citeable'] = True
+            self.obj['refereed'] = True
```
kaplun: @michamos can you double check this algorithm?

inspire_schemas/literature_builder.py
kaplun: This should actually be the `imprint` date which is specific to books (so `preprint_created` is wrong here.)

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+
+    def add_accelerator_experiment(self, experiment):
+        """
+        :param experiment: experiment correlated to the current artifact.
+        :type experiment: string
+        """
+        self._ensure_field('accelerator_experiments', [])
+
+        self.obj['accelerator_experiments'].append({
+            'experiment': experiment
```
kaplun: @michamos this should actually be `legacy_experiment` right?

inspire_schemas/literature_builder.py
```diff
@@ -0,0 +1,434 @@
...
+    def add_report_numbers(self, report_numbers):
+        """
+        :param report_numbers: report number for the current artifact.
+        :type report_numbers: list [{'report_number': '123'}].
+        """
+        self._ensure_field('report_numbers', [])
+
+        self.obj['report_numbers'].extend([{
+            'value': report_number.get('report_number', ''),
+        } for report_number in report_numbers])
```
kaplun: Here too you need to add the `source` information.

inspire_schemas/literature_builder.py
rikirenz: @michamos yes please check it. I have inherited this from the old code. I do not know if it is correct

david-caro:
Can you also import the LiteratureBuilder into the api module so you can just:
```python
from inspire_schemas.api import LiteratureBuilder
```

Something like:
```python
from .literature_builder import LiteratureBuilder  # noqa
```

Thanks!!

michamos:
No, it is wrong.

* Refereed is impossible to know based on only the article metadata.
Basically a paper is refereed if it has been published in a journal that we
consider to be doing serious peer review. It is more complicated than that
because some of those journals also publish conferences, and those are
usually not referred thoroughly.

* A paper is citeable if a reference to it can be matched with high
probability. This is the case if it has a complete publication_info
(journal, volume, page range or Id, and sometimes issue if the page range
is not unique among issues of the same volume) or a doi, or another
persistent identifier.

Both of these could be set automatically based on this policy, but they are
probably too complex to be handled by the literature builder.
They should also be overridable by the cataloguer on a case by case basis,
which we have no way to express in the current schema.

Il 10 feb 2017 6:55 PM, "Riccardo" <notifications@github.com> ha scritto:

*@rikirenz* commented on this pull request.
------------------------------

In inspire_schemas/literature_builder.py
<https://github.com/inspirehep/inspire-schemas/pull/81>:

> +                pass
+
+        self.obj['publication_info'].append(publication_item)
+
+        # Consideration about citeable and refereed attributes
+        has_pub_info = all([
+            key in self.obj['publication_info'] for key in (
+                'year', 'journal_issue', 'journal_volume')])
+
+        has_page_or_artid = any([
+            key in self.obj['publication_info'] for key in (
+                'page_start', 'page_end', 'artid')])
+
+        if has_pub_info and has_page_or_artid:
+            self.obj['citeable'] = True
+            self.obj['refereed'] = True

@michamos <https://github.com/michamos> yes please check it. I have
inherited this from the old code. I do not know if it is correct

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/inspirehep/inspire-schemas/pull/81>, or mute the thread

inspire_schemas/literature_builder.py
michamos: No, it is wrong.

* Refereed is impossible to know based on only the article metadata.
Basically a paper is refereed if it has been published in a journal that we
consider to be doing serious peer review. It is more complicated than that
because some of those journals also publish conferences, and those are
usually not referred thoroughly.

* A paper is citeable if a reference to it can be matched with high
probability. This is the case if it has a complete publication_info
(journal, volume, page range or Id, and sometimes issue if the page range
is not unique among issues of the same volume) or a doi, or another
persistent identifier.

Both of these could be set automatically based on this policy, but they are
probably too complex to be handled by the literature builder.
They should also be overridable by the cataloguer on a case by case basis,
which we have no way to express in the current schema.

inspire_schemas/literature_builder.py
rikirenz: Anyway I think this should not be done here. Who is collecting data should check this.

inspire_schemas/literature_builder.py
rikirenz: This means that the control that I put is valid for the `citeable` field but is invalid for the `refereed` field.

inspire_schemas/literature_builder.py
michamos: https://github.com/inspirehep/refextract/blob/master/refextract/references/regexs.py#L649

inspire_schemas/literature_builder.py
rikirenz: Of course this is not the best way to handle authors. Because there is not a specific format.

This way is the old way to handle Authors. 

IMO we have  2 alternatives because Author is a complex field and the format that the builder requires has to be as much strict as possible:

1. Create a specific builder for the author. This builder returns a `inspire_author` object and we add a method called `add_author()` to `literature_builder`.
2. We can define a method `add_author(param1,..., paramN)` this method has to create the author in according to our schema and adding it to the hep record that is generating.

Let me know which solution you prefer.

inspire_schemas/literature_builder.py
rikirenz: `legacy_experiment ` does not exist in the `hep` schema -> https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/hep.json

inspire_schemas/literature_builder.py
rikirenz: @michamos amazing!

inspire_schemas/literature_builder.py
michamos: [`legacy_name`](https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/hep.json#L42)

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+    def add_doi(self, doi, source=None):
+        """Add doi.
+
+        :param doi: doi for the current document.
+        :type doi: string
+
+        :param source: source for the current doi.
+        :type source: string
+        """
+        def _doi_normalization(doi_id):
```
kaplun: Can this normalizations be moved inside something such `inspire_schemas.utils` so that they can be re-used in many contextes.

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+                \d{4}                             # [0-9] x4
+                (/|%2f)                           # / (possibly urlencoded)
+                [\w\-_:;\(\)/\.<>]+               # any character
+                [\w\-_:;\(\)/<>])                 # any char except full stop
+                """, re.VERBOSE + re.IGNORECASE))
+            return pattern.match(doi_id)
+
+        self._ensure_field('dois', [])
+
+        if _doi_normalization(doi):
```
kaplun: Mmh... and actually this one is kind of error-prone. If the `doi` is not normalizable it will silently be ignored. I am not sure where it would be the best position to put this checks/normalization: @jacquerie @jmartinm @david-caro any thouhts?
The thing is that if the client is responsible for providing always normalized/correct data then we have to repeat this in all clients. Additionally normalization/validation is schema specific so it should sit in `inspire-schema`.

So... I believe `inspire-schemas` should centralize these rules, but it's client code that should call them before passing them to the builder... What do you think?

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+            self.obj['citeable'] = True
+
+    def add_authors(self, authors, supervisors=None):
+        """Add authors.
+
+        :param authors: authors for a given document.
+        :type authors: list
+
+        :param supervisors: list of supervisor for a given document.
+        :type supervisors: list
```
kaplun: Can you document that, either an author has to either be a `dict` with a given set of keys or it has to be created with the dedicated helper.

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+        """
+        def _match_authors_initials(author_name):
+            return not bool(re.compile(r'[^A-Z. ]').search(author_name))
+
+        def _author_name_with_initials(current_author):
```
kaplun: What is this function doing?

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+    def add_authors(self, authors, supervisors=None):
+        """Add authors.
+
+        :param authors: authors for a given document.
+        :type authors: list
+
+        :param supervisors: list of supervisor for a given document.
+        :type supervisors: list
+        """
+        def _match_authors_initials(author_name):
```
kaplun: What is this function doing?

inspire_schemas/api.py
kaplun: If necessary, please rename them so that it's clear to the reader.

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+                current_author['full_name'] = ', '.join(name)
+            return current_author
+
+        def _author_filter(current_author):
+            current_author = _author_name_with_initials(current_author)
+            if current_author.get('affiliation'):
+                current_author['affiliations'] = [dict(
+                    value=current_author['affiliation']
+                )]
+            del current_author['affiliation']
```
kaplun: This helper make me really thing we need a clean way to create an author in the builder :+1:

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+            return current_author
+
+        if supervisors is None:
+            supervisors = []
+
+        self._ensure_field('authors', [])
+
+        authors = filter(None, authors)
```
kaplun: Am I wrong or this does not do anything? (beside crashing in case `authors=None`)

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+        for author in authors:
+            author = _author_filter(author)
```
kaplun: What is the filter doing? It is not clear from its body or name...

inspire_schemas/api.py
```diff
@@ -57,3 +62,446 @@ def validate(data, schema_name=None):
...
+
+        :param date: publish date for the current thesis.
+        :type date: string. A date format is required (yyyy-mm-dd).
+        """
+        self._ensure_field('thesis_info', {})
+
+        thesis_item = {}
+        for key in ('defense_date', 'degree_type', 'date'):
+            if locals()[key] is not None:
+                thesis_item[key] = locals()[key].lower()
```
kaplun: Why this?

rikirenz: On suggestion of @david-caro is a way to avoid several `if` with the same logic.

inspire_schemas/api.py
rikirenz: I am going to rewrite this specific function

inspire_schemas/api.py
kaplun: It's quite cryptic though. It's quite some magic... But I guess it cannot be made any simpler.

inspire_schemas/api.py
rikirenz: Probably I miss understood the question. It is the filter function of python
it removes `None` objects if are present:
```
In[1]: a = [None,1,3]
In[2]: a
Out[2]: 
[None, 1, 3]
In[3]: a = filter(None, a)
In[4]: a
Out[4]: 
[1, 3]

```

kaplun: No here it was about: `_author_filter` :)

inspire_schemas/api.py
```diff
@@ -57,3 +62,468 @@ def validate(data, schema_name=None):
...
+        def _verify_author_name_initials(author_name):
+            return not bool(re.compile(r'[^A-Z. ]').search(author_name))
+
+        def _normalize_author_name_with_comma(author_name):
+            name = author_name.split(',')
+            if len(name) > 1 and _verify_author_name_initials(name[1]):
+                name[1] = name[1].replace(' ', '')
+            name = ', '.join(name)
+            return name
+
```
rikirenz: @kaplun This code is trying to simulate this behavior that was in the old code: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/dojson/fields/literature.py#L71-L83. Of course it needs tests. But please have a look and check if I am translating the right behavior

kaplun: I think this piece of code is one of those that should be factored out and put in the normalization utilities.

The meaning of this utils is:
In INSPIRE a name is presented as:
`Last names, First names`

where `First names` could be made of initials. I.e. by single letters followed by dots. In this case there should not be spaces.

inspire_schemas/api.py
rikirenz: @jacquerie @jmartinm @david-caro sorry guys but I would like to close this PR as soon as possible in order to have a stable builder. What do you think about the normalization? Who should do this?

michamos: you could use [idutils.normalize_doi](http://idutils.readthedocs.io/en/latest/#idutils.normalize_doi) instead.

inspire_schemas/api.py
```diff
@@ -59,3 +66,541 @@ def validate(data, schema_name=None):
...
+        self.obj['collaborations'].append({
+            'value': collaboration
+        })
+
+    def add_acquisition_source(self,
+                               method,
+                               submission_number=None,
+                               email=None,
+                               source=None,
+                               orcid=None):
```
kaplun: Can you also add the user id:
https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/elements/acquisition_source.json#L12

inspire_schemas/api.py
```diff
@@ -59,3 +66,541 @@ def validate(data, schema_name=None):
...
+
+        self.obj['urls'].append({
+            'value': url
+        })
+
+    def add_report_numbers(self, report_numbers, source=None):
+        """Add report numbers.
+
+        :param report_numbers: report number for the current document.
+        :type report_numbers: list [{'report_number': '123'}].
```
kaplun: Why not just a list of string?

kaplun: Or in fact this could follow the same pattern of the other fields and you would just accept one reportnumber at a time and let the client code do the looping.

inspire_schemas/api.py
```diff
@@ -59,3 +66,541 @@ def validate(data, schema_name=None):
...
+        :type language: string (2 characters)
+        """
+        self._ensure_field('languages', [])
+
+        self.obj['languages'].append(language)
+
+    def add_license(self, url, license=None):
+        """Add license url.
+
+        :param license_url: url for the description of the license.
```
kaplun: This could also be `None`

inspire_schemas/api.py
```diff
@@ -59,3 +66,541 @@ def validate(data, schema_name=None):
...
+        for key in ('defense_date', 'degree_type', 'date'):
+            if locals()[key] is not None:
+                thesis_item[key] = locals()[key].lower()
+
+        if institution is not None:
+            thesis_item['institutions'] = [{'name': institution}]
+
+        self.obj['thesis_info'] = thesis_item
+
+    def add_accelerator_experiment(self, experiment):
```
kaplun: The parameter should probably be really called `legacy_name` until we will have properly migrated to the split version.

inspire_schemas/api.py
```diff
@@ -59,3 +66,541 @@ def validate(data, schema_name=None):
...
+            'date': imprint_date
+        })
+
+    def add_preprint_date(self, preprint_date):
+        """Add imprint date.
+
+        :param preprint_date: preprint date.
+        :type preprint_date: string. A date format is required (yyyy-mm-dd).
+        """
+        self.obj['preprint_date'] = preprint_date
```
kaplun: I am not sure: is `yyyy-mm-dd` a valid jsonschema date?
https://tools.ietf.org/html/rfc3339#section-5.6
I fear we need to amend here and complete with `T00:00:00`

inspire_schemas/api.py
```diff
@@ -59,3 +66,615 @@ def validate(data, schema_name=None):
...
+    def _get_source(self, source):
+        if source is not None:
+            return source
+        return self.source
+
+    def add_if_non_empty(self, property, **kwargs):
+        """Add a property avoiding empty values.
+
+        Add a specific field to the record obj with all
+        the attributes that are not empty (None, '', [], {}).
```
kaplun: Can you add an example of usage?
e.g.

```python
add_if_non_empty('abstract', abstract='foo')
```

kaplun: BTW in English should it `add_if_noN_empty` or `add_if_noT_empty`. I am not actually sure.

inspire_schemas/api.py
```diff
@@ -59,3 +66,615 @@ def validate(data, schema_name=None):
...
+            if value in EMPTIES:
+                continue
+            mykwargs[key] = value
+
+        if _verify_empty_kwargs_without_source_key(copy(mykwargs)):
+            return
+
+        myfunc = getattr(self, 'add_' + property, None)
+        if myfunc is None:
+            raise Exception('Builder does not have method add_' + property)
```
kaplun: You could raise `RuntimeError` here, which is more appropriate.

rikirenz: Actually we should write a specific exception for this

inspire_schemas/api.py
rikirenz: you are right!

inspire_schemas/api.py
rikirenz: > `Non` is not a word, but a part of another word, usually a descriptive adjective
http://english.stackexchange.com/questions/277210/no-not-and-non

david-caro: I'd say non-empty (as in non-empty value).

inspire_schemas/api.py
david-caro: You can even make it inherit from the runtime error :)

david-caro: You can even make it inherit from the runtime error :)

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        this call is equivalent to add_abstract(abstract='foo')
+        2 - add_if_non_empty('abstract', abstract='')
+        this call is not going to execute any method
+
+        :param property: object property
+        :type property: string
+
+        :param **kwargs: List of parameters that a specific
+        property requires
+        :type **kwargs: {key: value}
```
jmartinm: The type would be `dict`

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        def _verify_empty_kwargs_without_source_key(curr_kwargs=None):
+            if curr_kwargs.get('source'):
+                del curr_kwargs['source']
+            if not curr_kwargs:
+                return True
+            return False
+
+        EMPTIES = [None, '', [], {}]
+        mykwargs = {}
+        for key, value in kwargs.items():
```
jmartinm: You could use a `dict` comprehension:

```python
mykwargs = {key: val for key, val in kwargs.items() if val not in EMPTIES}
```

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+            if value in EMPTIES:
+                continue
+            mykwargs[key] = value
+
+        if _verify_empty_kwargs_without_source_key(copy(mykwargs)):
+            return
+
+        myfunc = getattr(self, 'add_' + property, None)
```
jmartinm: `property` is a keyword in Python so better to use a different name.

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+            return False
+
+        EMPTIES = [None, '', [], {}]
+        mykwargs = {}
+        for key, value in kwargs.items():
+            if value in EMPTIES:
+                continue
+            mykwargs[key] = value
+
+        if _verify_empty_kwargs_without_source_key(copy(mykwargs)):
```
jmartinm: I am not sure I understand what this is for, can you explain it?

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        """Add author.
+
+        :param author: author for a given document.
+        :type author: object that comes
+        """
+        self._ensure_field('authors', [])
+
+        self.obj['authors'].append(author)
+
+    def make_author(self, full_name, affiliations=None, supervisor=False):
```
jmartinm: If there is no need for an instance to be available when calling this, you can decorate with `@staticmethod`

Then from the outside somebody can use it `LiteratureBuilder.make_author`

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        self.obj['copyright'].append(copyright)
+
+    def add_number_of_pages(self, number_of_pages):
+        """Add number_of_pages.
+
+        :param number_of_pages: Number of pages for
+        the current document
+        :type number_of_pages: int
+        """
+        self._ensure_field('number_of_pages', 1)
```
jmartinm: Do you need this?

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+
+        self.obj['publication_type'].append(publication_type)
+
+    def set_core(self, core):
+        """Set core value.
+
+        :param core: define a core article
+        :type core: bool
+        """
+        self._ensure_field('core', False)
```
jmartinm: Don't think you need this, you can directly assign the value.

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+
+        self.obj['core'] = core
+
+    def set_refereed(self, refereed):
+        """Set refereed value.
+
+        :param refereed: define a refereed article
+        :type refereed: bool
+        """
+        self._ensure_field('refereed', False)
```
jmartinm: same here, you can directly assign the value (in all cases when it is a primitive value)

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+
+            def __str__(self):
+                return repr('Builder does not have method add_' + self.method)
+
+        def _verify_empty_kwargs_without_source_key(curr_kwargs=None):
+            if curr_kwargs.get('source'):
+                del curr_kwargs['source']
+            if not curr_kwargs:
+                return True
+            return False
```
jacquerie: If I understand where you are going with this, all this method could just be
```python
return curr_kwargs.keys() == ['source']
```

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        class MissingMethodError(RuntimeError):
+            """Exception raised when a method is missing."""
+
+            def __init__(self, method):
+                self.method = method
+
+            def __str__(self):
+                return repr('Builder does not have method add_' + self.method)
+
+        def _verify_empty_kwargs_without_source_key(curr_kwargs=None):
```
jacquerie: Since this method returns a `bool` the Python convention is to call something like `is_...` or `has_...`, so that when you write this expression in an `if` statement it reads nicely. In this case, maybe
```python
def _has_only_source(d):
```

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        :type **kwargs: {key: value}
+        """
+        class MissingMethodError(RuntimeError):
+            """Exception raised when a method is missing."""
+
+            def __init__(self, method):
+                self.method = method
+
+            def __str__(self):
+                return repr('Builder does not have method add_' + self.method)
```
jacquerie: Exceptions that your code can raise are part of the API, so they should live at the top level (otherwise you can't import them to catch them).

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+            if locals()[key] is not None:
+                acquisition_source[key] = locals()[key]
+
+        self.obj['acquisition_source'] = acquisition_source
+
+    def add_document_type(self, document_type):
+        """Add document type.
+
+        :param document_type: document type.
+        :type document_type: string
```
jacquerie: This is not any string, but a string in a restricted set of values. If the builder does not ensure this, then who is going to do it?

inspire_schemas/utils.py
```diff
@@ -108,3 +110,39 @@ def load_schema(schema_name):
...
+
+    :param date: a generic date
+    :type date: string with the format (yyyy-mm-dd)
+
+    :return formatted_date: the input date in
+    the format (yyyy-mm-ddT00:00:00)
+    """
+    try:
+        formatted_date = datetime.datetime.strptime(date, '%Y-%m-%d')
+        formatted_date = formatted_date.isoformat().lower()
```
jacquerie: Why `lower`?

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        for key in ('defense_date', 'degree_type', 'date'):
+            if locals()[key] is not None:
+                thesis_item[key] = locals()[key].lower()
+
+        if institution is not None:
+            thesis_item['institutions'] = [{'name': institution}]
+
+        self.obj['thesis_info'] = thesis_item
+
+    def add_legacy_name(self, legacy_name):
```
jmartinm: `add_legacy_name` seems too generic name since it is related to `accelerator_experiments`, you can name it `add_ accelerator_experiments_legacy_name`

inspire_schemas/api.py
michamos: @david-caro says:

> No :smiley:

inspire_schemas/utils.py
```diff
@@ -108,3 +110,39 @@ def load_schema(schema_name):
...
+def normalize_author_name_with_comma(author):
+    """Normalize author name.
+
+    :param author: a generic date
+    :type author: string
+
+    :return name: the name of the author normilized
+    """
+    def _verify_author_name_initials(author_name):
+        return not bool(re.compile(r'[^A-Z. ]').search(author_name))
```
michamos: * What about non-ASCII initials?
* The name of the function could be more explicit.

inspire_schemas/api.py
rikirenz: We are checking that at least one parameter is present when we call a specific method, except for the `source` field.
If all the parameters are empty and the `source` is a constant like _arXiv_; the method is going to be invoked anyway if we consider it as a normal parameter. We should not consider the `source` because an object that has a `source` but it does not have all the other values is not useful.

inspire_schemas/api.py
rikirenz: ok

inspire_schemas/api.py
rikirenz: ok for both!

inspire_schemas/utils.py
rikirenz: Well all the schema is lower case. I thought let's do also the iso_format lower case and it also not conflicting with the ISO_8601 standard (http://dotat.at/tmp/ISO_8601-2004_E.pdf). Anyway if you want I can leave it upper case.

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
         resolver=LocalRefResolver.from_schema(schema),
         format_checker=draft4_format_checker,
     )
+
+
+class LiteratureBuilder(object):
+    """Literature Suggestion Builder."""
+
+    def __init__(self, source):
```
rikirenz: I think we can do it. It should not create problems :smiley:

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        this call is not going to execute any method
+
+        :param property: object property
+        :type property: string
+
+        :param **kwargs: List of parameters that a specific
+        property requires
+        :type **kwargs: {key: value}
+        """
+        class MissingMethodError(RuntimeError):
```
rikirenz: Well I do not have a good question to justify my code in according to your comment. If also @david-caro  and @kaplun do not have justifications. I am going to use `AttributeError`

inspire_schemas/api.py
```diff
@@ -59,3 +66,633 @@ def validate(data, schema_name=None):
...
+        :param arxiv_categories: arXiv categories for the current document.
+        :type arxiv_categories: list
+        """
+        self._ensure_field('arxiv_eprints', [])
+
+        self.obj['arxiv_eprints'].append({
+            'value': arxiv_id,
+            'categories': arxiv_categories,
+        })
+        self.set_citeable(True)
```
rikirenz: This make sense but IMHO this consideration should be put in the `inspire-schema`. 
Something like this:
``` python
    record_with_citeable_field_setted = set_citeable_field(record)
```

inspire_schemas/api.py
jacquerie: We discussed this with @david-caro, so don't worry about this.

inspire_schemas/utils.py
michamos: The only letter is the `T` in the middle, which is usually uppercase.

jacquerie: Apologies for language lawyering.

It all depends on what `format_checker` we use. If it's the default one in `jsonschema`, without any extra dependencies installed (https://github.com/Julian/jsonschema/blob/7ccd4ff376f8417a71673b7c761e65a9ee8a0bef/jsonschema/_format.py#L205-L223) then we validate as ISO 8601, which allows for lowercase characters. If instead we validate as RFC3339 then it's invalid (read the grammar if you don't believe me: https://tools.ietf.org/html/rfc3339#page-12).

rikirenz: I will leave upper case in order to maintain the compatibly for both.

inspire_schemas/api.py
rikirenz: I am going to change this in order to pass a list

inspire_schemas/utils.py
david-caro: Let's put that improvement on an issue, this is some legacy logic that 'somewhat' worked so far, I agree that it needs rethinking and maybe some refactoring, but I don't think this pr is the right place as this is kind of unrelated to the builder itself.

rikirenz: I am going to create the issue :+1:

inspire_schemas/builders.py
```diff
@@ -0,0 +1,684 @@
...
+    @wraps(func)
+    def func_wrapper(self, *args, **kwargs):
+        empties = [None, '', [], {}]
+        my_kwargs = {}
+        for key, value in kwargs.items():
+            if value in empties:
+                continue
+            my_kwargs[key] = value
+
+        if (my_kwargs.keys() == ['source'] or my_kwargs.keys() == []) and args == ():
```
michamos: this fails on Python 3 (which we target), as keys() returns a dictionary view rather than a list. You should wrap it in `list()`.

inspire_schemas/builders.py
```diff
@@ -0,0 +1,684 @@
...
+    """Decorator that is filtering empty parameters.
+
+    :param func: function that you want wrapping
+    :type func: function
+    """
+    @wraps(func)
+    def func_wrapper(self, *args, **kwargs):
+        empties = [None, '', [], {}]
+        my_kwargs = {}
+        for key, value in kwargs.items():
```
michamos: `my_kwargs = {key: value for key, value in kwargs.items() if value not in empties}` is probably more readable

inspire_schemas/builders.py
```diff
@@ -0,0 +1,684 @@
...
+        self.record = {} if old_record is None else old_record
+        self.source = source
+
+    def __str__(self):
+        """Print the current record."""
+        return str(self.record)
+
+    def __repr__(self):
+        """Printable representation of the builder."""
+        return '<source:%s, obj:%s>' % (self.source, self.record)
```
michamos: `return 'LiteratureBuilder(`...`)'` as I mentioned before

inspire_schemas/builders.py
```diff
@@ -0,0 +1,690 @@
...
+        :param arxiv_categories: arXiv categories for the current document.
+        :type arxiv_categories: list
+        """
+        self._ensure_field('arxiv_eprints', [])
+
+        self.record['arxiv_eprints'].append({
+            'value': arxiv_id,
+            'categories': arxiv_categories,
+        })
+        self.set_citeable(True)
```
michamos: We keep it like this in the end?

inspire_schemas/builders.py
```diff
@@ -0,0 +1,690 @@
...
+        :param roles: it tells the roles of
+        the current author.
+        :type roles: list
+
+        :return: dict {fullname:}
+        """
+        def _add_affiliations(author, affiliations):
+            for affiliation in affiliations:
+                if affiliation:
+                    ensure_field('affiliations', [], author)
```
michamos: you can take this out of the `for` loop

inspire_schemas/builders.py
```diff
@@ -0,0 +1,690 @@
...
+                self.add_number_of_pages(
+                    int(page_end) - int(page_start) + 1
+                )
+            except (TypeError, ValueError):
+                pass
+
+        self.record['publication_info'].append(publication_item)
+
+        # Consideration about citeable attribute
+        has_pub_info = all([
```
michamos: you don't need the  surrounding `[ ]`, `all` works on any iterable

inspire_schemas/builders.py
```diff
@@ -0,0 +1,690 @@
...
+            key in self.record['publication_info'] for key in (
+                'year', 'journal_issue', 'journal_volume'
+            )
+        ])
+
+        has_page_or_artid = any([
```
michamos: same as before

inspire_schemas/builders.py
```diff
@@ -0,0 +1,690 @@
...
+            key in self.record['publication_info'] for key in (
+                'page_start', 'page_end', 'artid'
+            )
+        ])
+
+        if has_pub_info and has_page_or_artid:
+            self.set_citeable(True)
+
+    @filter_empty_parameters
+    def add_imprint(self, imprint_date):
```
michamos: `add_imprint_date`

inspire_schemas/builders.py
```diff
@@ -0,0 +1,690 @@
...
+
+    def _ensure_field(self, field_name, value):
+        ensure_field(field_name, value, self.record)
+
+    def _get_source(self, source):
+        if source is not None:
+            return source
+        return self.source
+
+    def validate_object(self):
```
michamos: `validate_record`

inspire_schemas/builders.py
```diff
@@ -0,0 +1,676 @@
...
+            return
+        return func(self, *args, **my_kwargs)
+
+    return func_wrapper
+
+
+class LiteratureBuilder(object):
+    """Literature record builder."""
+
+    def __init__(self, source, old_record=None):
```
jmartinm: I would just call it `record` instead of `old_record`

inspire_schemas/builders.py
david-caro: Yep, to be fixed after discussion on #103

inspire_schemas/builders.py
```diff
@@ -0,0 +1,676 @@
...
+
+    def __init__(self, source, record=None):
+        """Init method.
+
+        :param source: sets the default value for the 'source' fields
+         of the current record, which captures where the information
+         that the builder populates comes from
+        :type source: string
+
+        :param old_record: sets the default value for the the current
```
michamos: you didn't update the docstring here

inspire_schemas/builders.py
```diff
@@ -0,0 +1,676 @@
...
+        record, in order to edit an already existent record
+        :type old_record: dict
```
michamos: same comment

inspire_schemas/builders.py
```diff
@@ -0,0 +1,676 @@
...
+        self.record['accelerator_experiments'].append({
+            'legacy_name': legacy_name
+        })
+
+    @filter_empty_parameters
+    def add_language(self, language):
+        """Add language.
+
+        :param language: language for the current document
+        :type language: string (2 characters ISO639-2)
```
michamos: we use ISO 639-1, not -2.

inspire_schemas/builders.py
```diff
@@ -0,0 +1,676 @@
...
+
+    @filter_empty_parameters
+    def add_title_translation(self, title, language=None, source=None):
+        """Add title translation.
+
+        :param title: translated title
+        :type title: string
+
+        :param language: language for the original title
+        :type language: string (2 characters ISO639-2)
```
michamos: same comment

inspire_schemas/builders.py
```diff
@@ -0,0 +1,676 @@
...
+    def add_abstract(self, abstract, source=None):
+        """Add abstract.
+
+        :param abstract: abstract for the current document.
+        :type abstract: string
+
+        :param source: source for the given abstract.
+        :type source: string
+        """
+        self._ensure_field('abstracts', [])
```
jmartinm: I am not sure about so many indirections of calling `_ensure_field` that then calls `ensure_field` since it can be done here with a oneliner:

```python
self.record.set_default('abstracts', [])
```

And it makes it more readable instead of having to check each of those functions.

inspire_schemas/builders.py
```diff
@@ -0,0 +1,676 @@
...
+        self.record = {} if record is None else record
+        self.source = source
+
+    def __str__(self):
+        """Print the current record."""
+        return str(self.record)
+
+    def __repr__(self):
+        """Printable representation of the builder."""
+        return 'LiteratureBuilder(source="{}", old_record={})'.format(
```
jmartinm: leftover `old_record` -> `record`

tests/integration/test_builders.py
```diff
@@ -0,0 +1,145 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE-SCHEMAS.
+# Copyright (C) 2016 CERN.
```
jmartinm: 2017

PR #83, opened 2017-02-13
schemas: strictness changes some fields

michamos:
* INCOMPATIBLE All 'value's are now required.
* The ISBN pattern has been relaxed, as there are invalid ISBNs in the
nature, which we still want to keep track of.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -105,6 +105,9 @@
                                     "type": "string"
                                 }
                             },
+                            "required": [
+                                "value"
+                            ],
                             "title": "Affiliation",
                             "type": "object"
                         },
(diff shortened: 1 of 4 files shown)


inspire_schemas/records/hep.json
```diff
@@ -373,6 +379,10 @@
                         "type": "string"
                     }
                 },
+                "required": [
+                    "schema",
```
kaplun: Mmh... This is true, we need then to check how many IDs on legacy don't have a schema and would therefore need to be either guessed by dojson or fixed on legacy.

michamos: We have 57 of them, correcting right now on legacy.

PR #86, opened 2017-02-15
schemas: changes in acquisition_source

michamos:
* INCOMPATIBLE Uses for `method` the enum that was previously used for
`source`. `source` is for where the information comes from, `method` for
how we obtained it.
* `source` uses the source element.
* Adds `internal_uid` key for the user identifiers on the server.

### DIFF
--- a/inspire_schemas/records/elements/acquisition_source.json
+++ b/inspire_schemas/records/elements/acquisition_source.json
@@ -6,16 +6,13 @@
             "type": "string"
         },
         "email": {
+            "format": "email",
             "type": "string"
+        },
+        "internal_uid": {
+            "type": "integer"
         },
         "method": {
-            "type": "string"
-        },
-        "orcid": {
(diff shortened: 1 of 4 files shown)


michamos:
relevant for @jacquerie and @rikirenz

inspire_schemas/records/elements/acquisition_source.json
```diff
@@ -8,14 +8,10 @@
         "email": {
             "type": "string"
         },
-        "method": {
-            "type": "string"
+        "internal_uid": {
```
kaplun: Can you add above the native format `email`?

inspire_schemas/records/elements/acquisition_source.json
```diff
@@ -8,14 +8,10 @@
         "email": {
             "type": "string"
```
kaplun: Can you add above the native format email?

PR #88, opened 2017-02-15
hep: new _fft field

michamos:
* Adds an _fft field, with contents based on
https://github.com/inspirehep/inspire/blob/master/bibformat/format_elements/bfe_INSPIRE_enhanced_marcxml.py#L152-L164.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -13,6 +13,43 @@
             },
             "type": "array"
         },
+        "_fft": {
+            "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


michamos:
I believe only HEP records can have FFTs, is that correct @kaplun?

jacquerie:
Err... was this really correct? A record can have multiple `FFT`s. For example: https://inspirehep.net/record/863300/export/xme

kaplun:
Agh! Sure! I should not merge things after dinner...

PR #91, opened 2017-02-16
hep: publication_type is an array

michamos:
* INCOMPATIBLE makes `publication_type` and array.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -708,12 +708,15 @@
             "uniqueItems": true
         },
         "publication_type": {
-            "enum": [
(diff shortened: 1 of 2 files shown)


jacquerie:
Uh. For some reason this generated `inspire-schemas==18.0.3` instead of `inspire-schemas==20.0.0`: https://pypi.python.org/pypi/inspire-schemas/18.0.3

@david-caro, any idea why?

david-caro:
Hmm, maybe on travis it's not downloading the whole got log? Of so, we can bypass it with tags (create a tag for the latest major, we have pending automating the tag creation...)

PR #93, opened 2017-02-17
elements: changes in id

michamos:
* INCOMPATIBLE `type` is renamed to `schema` for consistency.
* Adds old `SPIRES` id.


michamos:
Note that this does not solve the problem for the other collections which are still missing some place to store the SPIRES identifier.

PR #97, opened 2017-02-21
Schema documentation

michamos:
Documentation is mostly done for HEP and the elements used in it. There are some small display bugs, but any comments would already be very appreciated. A typeset version of the schemas (using @ammirate's `yaml2rst`) can be found at https://michamos.github.io/inspire-schemas/.


david-caro:
I really think we should try to do something about the multiline... and imo the yaml format is still the best alternative (even with the required parsing in order to get a json...).

inspire_schemas/records/elements/acquisition_source.json
```diff
@@ -22,13 +27,15 @@
             "type": "string"
         },
         "orcid": {
+            "description": "'If `method` is `submitter`, stores the ORCID of the submitter.\n\n:MARC: ``541__a`` with ``orcid:`` prefix\n",
```
david-caro: What happens if not?

inspire_schemas/records/elements/acquisition_source.json
```diff
@@ -22,13 +27,15 @@
...
             "pattern": "^\\d{4}-\\d{4}-\\d{4}-\\d{3}[0-9X]$",
             "type": "string"
         },
         "source": {
             "$ref": "source.json"
         },
         "submission_number": {
+            "description": "If `method` is `submitter`, stores the holding pen record ID of the submission.\n\n:MARC: ``541__e``\n",
```
david-caro: same here

inspire_schemas/records/elements/id.yml
```diff
@@ -3,6 +3,12 @@
     "anyOf": [
         {
             "additionalProperties": false,
+            "title": "Inspire ID",
+            "description": "
+                This identifier is mainly used by large collaborations
+                (providing an `authors.xml` file) to uniquely identify the
+                authors of their articles.
```
kaplun: So currently this ID is minted by us (actually by Margareth(?)) and in principle is issued for any curated HEPName.

inspire_schemas/records/elements/id.yml
```diff
@@ -23,6 +30,23 @@
         },
         {
             "additionalProperties": false,
+            "title": "Inspire BAI",
+            "description": "
+                The BAI (BibAuthorID) is based on the name of the author and is
+                used as a human-friendly unique identifier.
+                Its general form is a ``firstnames.lastnames.number``, where
+                ``firstnames`` is a list of first names (or initials) separated
```
kaplun: I would stress that by default we use initials and we fall-back on first names upon user request. At least this is the current policy.

inspire_schemas/records/elements/id.yml
```diff
@@ -23,6 +30,23 @@
...
+                by dots, and similarly for ``lastnames``. Only ASCII letters
+                (without diacritics), numbers, dashes and apostrophes are
+                allowed in the names.
+
+
+                .. note::
+
+                This identifier is not persistent, as it can be renamed to
+                become more explicit (e.g. transforming an initial into a full
+                name), or disappear when merging two different authors.
```
kaplun: This makes me think we should preserve previous ID into an obsolete list, so that we are always able to resolve them.

inspire_schemas/records/elements/id.yml
```diff
@@ -103,6 +152,7 @@
         },
         {
             "additionalProperties": false,
+            "title": "arXiv author identifier",
```
kaplun: More info here: https://arxiv.org/help/author_identifiers

inspire_schemas/records/elements/id.yml
```diff
@@ -123,6 +174,7 @@
         },
         {
             "additionalProperties": false,
+            "title": "CERN CCID",
```
kaplun: CERN Institutional ID?

inspire_schemas/records/elements/id.yml
```diff
@@ -163,6 +218,7 @@
         },
         {
             "additionalProperties": false,
+            "title": "Google Scholar identifier",
```
kaplun: Google Scholar profile identifier

annetteholtkamp:
It’s usually not by user request but at the discretion of the cataloguers. For very common names we would have many BAIs just differing by the appended number, e.g. for many Chinese names. In that case we usually decide to use full first names to make the BAIs better distinguishable.

- Annette


On 28 Feb 2017, at 16:51, Samuele Kaplun <notifications@github.com<mailto:notifications@github.com>> wrote:


@kaplun commented on this pull request.

________________________________

In inspire_schemas/records/elements/id.yml<https://github.com/inspirehep/inspire-schemas/pull/97#discussion_r103484973>:

> @@ -23,6 +30,23 @@
         },
         {
             "additionalProperties": false,
+            "title": "Inspire BAI",
+            "description": "
+                The BAI (BibAuthorID) is based on the name of the author and is
+                used as a human-friendly unique identifier.
+                Its general form is a ``firstnames.lastnames.number``, where
+                ``firstnames`` is a list of first names (or initials) separated


I would stress that by default we use initials and we fall-back on first names upon user request. At least this is the current policy.

michamos:
@inspirehep/inspire-content

inspire_schemas/records/elements/id.yml
hoc3426: Actually we use first names regularly to avoid ambiguity and make it easier to immediately know who is who. If the profile is James.Smith.1 it's immediately obvious that a John Smith doesn't belong. The full name also is a sign that someone has cleaned (to some extent) the profile.

michamos: Thanks for the comment Heath. Actually, this comment is quite old and referred to a previous version. For the latest changes, you can look at the attached files, or, better, the HTML output in https://michamos.github.io/inspire-schemas. For BAI, the explanation is here https://michamos.github.io/inspire-schemas/id.html#inspire-bai. Do you think it should be amended?

hoc3426:
Does experiments have a good place for these category codes? https://inspirehep.net/info/Experiments/list

michamos:
@hoc3426 we have not payed much attention to the secondary collections. For now, we have mainly worked HEP and fields that exist everywhere (`deleted`, etc.).

Are the codes you are referring to documented in the Twiki?

michamos:
can someone review this so it can be merged ASAP? given that the move to non-flow style YAML has completely rewritten everything, it will be a pain to maintain if the schema changes.

michamos:
I rebased on master to incorporate the changes introduced by #125.

For the record, this is the best way to do it to avoid huge conflicts is to use the `-X theirs` option of `git rebase` to avoid huge merge conflicts (when rebasing, theirs is confusingly the current branch, not master) and edit the modified file(s) (which have not changed after the rebase) to incorporate the changes.

PR #104, opened 2017-02-22
tests: remove useless pydocstyle

david-caro:
Also forcing a new release with the correct version. This commit might
be reverted after if so decided in the documentation meeting.


david-caro:
Merging this to get the correct version on pypi

PR #105, opened 2017-02-23
Stricter schemas

michamos:


inspire_schemas/records/elements/acquisition_source.json
```diff
@@ -2,7 +2,8 @@
     "$schema": "http://json-schema.org/schema#",
     "additionalProperties": false,
     "properties": {
-        "date": {
+        "datetime": {
+            "format": "date-time",
```
jacquerie: Do we want to convert each field in a separate INCOMPATIBLE commit, or should we do one big breaking commit?

michamos: I don't think there are many to do, all the others are supposed to be dates with no time.

inspire_schemas/records/authors.json
```diff
@@ -10,12 +10,15 @@
...
-            "type": "array"
+            "minItems": 1,
+            "type": "array",
+            "uniqueItems": true
         },
         "_private_notes": {
             "items": {
                 "$ref": "elements/sourced_value.json"
             },
+            "minItems": 1,
```
kaplun: Doesn't this means that there should always be a `_private_notes` item?

kaplun: Right. If there is `_private_notes` then it must have at least one item. Since we are cleaning up records recursively in `dojson` and we should do it also in the `builder` these requirements should be enforced already at application level, but better safe than sorry.

michamos: Exactly. If it's not enforced by the schema, then it cannot be assumed by any code that uses the schema. This should lead to some simplification.

jacquerie:
@michamos, is it ok for you if we merge 3fbac012d4b9416c1d68331cdb6b298de758a19b independently of this PR? I want to annotate in the mappings all the fields that contain dates with the right format, and I'd rather not have to modify it later...

jacquerie:
Actually, same goes for 57286f1ee7e797852e3a8ba1d3fb8db3d9497ca4. I think we learned what happens when we have to do too many schema version upgrades at the same time... : )

michamos:
If you want to, that's fine for me

jacquerie:
:+1:

This is the direction in which our schemas should go: strict schemas ensure that we don't have to deal in the application code with things like format of the input.

That said, we should deploy before merging this, so that if last-minute changes to the schemas are needed they are not blocked by potential problems given by the stricter schemas.

david-caro:
Is this PR still relevant?

michamos:
yes, we decided with @jacquerie that instead of splitting it into #255 and this one, we would do this once, so I incorporated those changes here.

michamos:
For review, it's best to look at the first commit and the script in the second one, as all the changes to the schemas there are produced by the script.

michamos:
@jacquerie do you want to review this before merge?

jacquerie:
> @jacquerie do you want to review this before merge?

I trust you :100: % (I reserve the right to complain afterwards.)

kaplun:
Move fast and break things :+1:

jacquerie:
> break things

ಠ_ಠ

PR #106, opened 2017-02-25
elements: expand regexp for BAIs

jacquerie:
```python
In [1]: schema = {'pattern': '^((\w|\-|\')+\.)+\d+$', 'type': 'string'}

In [2]: validate('L.Dell\'Asta.1', schema)

In [3]: validate('T.J.Berners-Lee.1', schema)

In [4]: with open('035__a_when_9_is_BAI.txt', 'r') as f:
   ...:     for line in f:
   ...:         try:
   ...:             validate(line.strip(), schema)
   ...:         except:
   ...:             print(line.strip())
```
prints
```
0000-0002-2836-6085
INSPIRE-00627028
INSPIRE-00584874
.J.M.Snellings.1
Qing.Ping.Ji.
Qi.Fang.Lü.1
10115251
.F.Schonfeld.1
```
where `035__a_when_9_is_BAI.txt` is a file containing all the BAIs I could find in HEPNames. These outliers could be interesting for @inspirehep/inspire-content. We don't want to allow characters like "ü" in BAIs, right?

### DIFF
--- a/inspire_schemas/records/elements/id.json
+++ b/inspire_schemas/records/elements/id.json
@@ -31,7 +31,7 @@
                     "type": "string"
                 },
                 "value": {
-                    "pattern": "^(\\w+\\.)+\\d+$",
+                    "pattern": "^((\\w|\\-|\\')+\\.)+\\d+$",
                     "type": "string"
                 }
             },


kaplun:
Correct. BAIs should be ASCII only.

michamos:
https://github.com/inspirehep/inspire-schemas/pull/106#issuecomment-282671828
> Correct. BAIs should be ASCII only.

OK, I will fix those from Jacopo's list then.

kaplun:
@jacquerie you can also find a list of BAIs as known in author profile via:
```sql
select distinct data from aidPERSONIDDATA where tag='canonical_name';
```

michamos:
> OK, I will fix those from @jacquerie's list then.

Done

kaplun:
This is the list from the profiles instead:
```
['.187', '.3', 'Cheng.Cai..Zhang.1', 'Guang.Lei.Li.1 ', '.1', 'M.L/Wagman.1', '.2', '.4', '.5', '.6', '.7', '.8', 'Pei.long.Wang.1 ']
```

Each one is reachable e.g.  through: http://inspirehep.net/author/profile/.187

PR #110, opened 2017-02-27
global: allow yaml schema files

david-caro:
* This will autogenerate the json files from any .yaml schema files at
  build time, allowing to use the more readable yaml format.


david-caro:
Of course, there aren't any yaml files yet to build here, but once we replace any .json with a .yaml, it will transparently do the transformation.

david-caro:
Finally passing all the tests :), @michamos, you can base your pr on this one.

michamos:
@david-caro Thanks, it is generating the JSON when running the tests but it is not prettified. Could you add a call to the prettifier?

david-caro:
> @david-caro Thanks, it is generating the JSON when running the tests but it is not prettified. Could you add a call to the prettifier?

Where is that prettifier?

david-caro:
Is `json.dumps(data, indent=4, sort_keys=True)` enough?

david-caro:
@michamos Can you rebase again into this one and try again? Now the json output should be already pretty :)

michamos:
I already fixed it in 633f392. Sorry for not letting you know.
Your gitignore changes are wrong, they are also ignoring the test fixtures.

david-caro:
You should rebase at least on top of all but the last two commits then, there's a fix on the test script (it was looking for .yaml, but we use .yml for the files extension).

jacquerie:
This was merged as part of #97.

PR #112, opened 2017-02-28
builders: fix publication_info citeable logic

rikirenz:
* Fixes the logic about citeable articles. Before it was considering
the publication_info field as an object rather than an array.

Signed-off-by: Riccardo Candido <riccardo.candido@gmail.com>

### DIFF
--- a/inspire_schemas/builders.py
+++ b/inspire_schemas/builders.py
@@ -144,7 +144,6 @@ class LiteratureBuilder(object):
                 'value': doi,
                 'source': self._get_source(source),
             })
-            self.set_citeable(True)
 
     @filter_empty_parameters
     def add_author(self, author):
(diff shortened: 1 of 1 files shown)


inspire_schemas/builders.py
```diff
@@ -290,16 +290,20 @@ def add_publication_info(
...
-        has_pub_info = all(
-            key in self.record['publication_info'] for key in (
-                'year', 'journal_issue', 'journal_volume'
-            )
+        has_pub_info = any(
+            all(
+                key in item for key in (
+                    'year', 'journal_issue', 'journal_volume'
+                )
+            ) for item in self.record['publication_info']
```
david-caro:
```python

def _is_citable(pub_info):
    def _item_has_pub_info(item):
        return all(
            key in item for key in ('year', 'journal_issue', 'journal_volume')
        )

    has_pub_info = any(_item_has_pub_info(item) for item in self.record['publication_info'])
    ...
    return has_pub_info and has_page_or_artid

if _is_citable(self.record['publication_info']):
    self.set_citable(True)
```

inspire_schemas/builders.py
```diff
@@ -270,6 +270,31 @@ def add_publication_info(
         the document has been published
         :type journal_volume: string
         """
+        def _is_citeable(publication_info):
+            # Consideration about citeable attribute
```
michamos: Could you maybe turn this into a proper docstring?

david-caro:
I'd say that first it needs more integration tests, and then unit tests. I would not be happy just with unit (and integration ones should be broader in scope than unit).

michamos:
after discussing with @annetteholtkamp this morning, turns out we don't consider DOI as citeable, but only complete publication info or arXiv. Could you remove the `set_citeable` from DOI while you're at it?

rikirenz:
@michamos could you have a look again now should be fine.

inspire_schemas/builders.py
```diff
@@ -270,6 +269,36 @@ def add_publication_info(
...
+            """Check some fields in order to define if the article is citeable.
+
+            :param publication_info: publication_info field
+            already populated
+            :type publication_info: list
+            """
+            def _item_has_pub_info(item):
+                return all(
+                    key in item for key in (
+                        'year', 'journal_issue', 'journal_volume'
```
michamos:
```
'year', 'journal_title', 'journal_volume'
```

inspire_schemas/builders.py
```diff
@@ -270,6 +269,36 @@ def add_publication_info(
...
+                    )
+                )
+
+            def _item_has_page_or_artid(item):
+                return any(
+                    key in item for key in (
+                        'page_start', 'page_end', 'artid'
```
michamos: no `page_end` here

rikirenz:
@michamos Should be fine now

PR #113, opened 2017-03-01
hep: add send_back flag to _fft

michamos:
* This  flag is used to decide whether to produce a `FFT` MARC field on
legacy from the `_fft` key in the record. This is necessary to avoid
sending FFTs back to legacy corresponding to files that are already
attached to the record, which would result in attaching them twice.

### DIFF
--- a/inspire_schemas/records/hep.json
+++ b/inspire_schemas/records/hep.json
@@ -72,6 +72,9 @@
                     "path": {
                         "type": "string"
                     },
+                    "send_back": {
+                        "type": "boolean"
+                    },
                     "status": {
                         "type": "string"
                     },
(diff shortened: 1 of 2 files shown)


kaplun:
Mmh. I am not sure I get the solution you propose, but I believe the problem is wider, and should be addressed in the general workflows of how things are synced between legacy and labs.

Actually the FFT tag should not even exist in the schema. 

We shall discuss handling fulltextes in general in one of the next technical meetings.

michamos:
yeah, this is an old one and was never merged because the right solution is probably to attach the files during the migration instead of writing the `_fft` tag. Let's close this.

PR #123, opened 2017-03-14
builders: fix add_acquisition_source

jacquerie:
Should have been part of #122.

### DIFF
--- a/inspire_schemas/builders.py
+++ b/inspire_schemas/builders.py
@@ -26,6 +26,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import warnings
 from functools import wraps
 import idutils
 
@@ -534,7 +535,8 @@ class LiteratureBuilder(object):
         internal_uid=None,
         email=None,
         orcid=None,
-        source=None
+        source=None,
+        datetime=None,
     ):
         """Add acquisition source.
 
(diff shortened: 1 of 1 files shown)


jacquerie:
This does not look right, because it's modifying the API of the builder, so all callers will have to amend their calls. Or is this how the builder is expected to be used? @david-caro

jacquerie:
CC: @david-caro @michamos

inspire_schemas/builders.py
```diff
@@ -556,13 +561,17 @@ def add_acquisition_source(
         :param internal_uid: id of the user that is creating the record
         :type internal_uid: string
         """
+        if date is not None:
+            warnings.warn('use datetime', DeprecationWarning)
```
michamos: I would add here
```
if datetime is not None:
  raise ValueError("Conflicting arguments given: 'date' and 'datetime'")
```

inspire_schemas/builders.py
```diff
@@ -530,6 +531,7 @@ def add_acquisition_source(
         self,
         method,
         date=None,
+        datetime=None,
```
michamos: You should probably add `datetime` at the end, otherwise it changes the function when called with positional (non-keyword) arguments.

jacquerie:
Both good points. Implemented them.

PR #124, opened 2017-03-14
added functions for book addition in the builder

Glignos:


inspire_schemas/builders.py
```diff
@@ -226,6 +226,75 @@ def _add_affiliations(author, affiliations):
...
+        """
+
+        imprint= {}
+
+        imprint['date']= date
+        imprint['place']= place
+        imprint['publisher']= publisher
+        self.record.setdefault('imprints', [])
+
+        self.record['imprints'].append(imprint)
```
jacquerie: Or just:
```python
self.record.setdefault('imprints', [])
self.record['imprints'].append({
    'date': date,
    'place': place,
    'publisher': publisher,
})
```

inspire_schemas/builders.py
```diff
@@ -226,6 +226,75 @@ def _add_affiliations(author, affiliations):
         return author
 
     @filter_empty_parameters
+    def add_book(self,publisher=None, place=None, date=None):
+        """
+        Make a dictionary that is representing a book.
```
jacquerie: Read the PEP257 on how to format docstrings: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

inspire_schemas/builders.py
```diff
@@ -226,6 +226,75 @@ def _add_affiliations(author, affiliations):
...
+        self.record.setdefault('imprints', [])
+
+        self.record['imprints'].append(imprint)
+
+    @filter_empty_parameters
+    def add_isbns(self,isbn):
+        """
+        :param isbns: the isbns of the book
+        :type isbns: list
+        """
```
jacquerie: What is it, `isbn` or `isbns`? It looks like here you changed your idea while you were writing this.

inspire_schemas/builders.py
jacquerie: (Same comment applies to `add_isbns` and `add_book_series`)

inspire_schemas/builders.py
```diff
@@ -331,7 +400,7 @@ def add_imprint_date(self, imprint_date):
         self.record.setdefault('imprints', [])
 
         self.record['imprints'].append({
-            'date': normalize_date_iso(imprint_date)
+            'date': imprint_date
```
jacquerie: Why this change? Looks unrelated.

inspire_schemas/builders.py
```diff
@@ -226,6 +226,75 @@ def _add_affiliations(author, affiliations):
         return author
 
     @filter_empty_parameters
+    def add_book(self,publisher=None, place=None, date=None):
+        """
+        Make a dictionary that is representing a book.
+
+        :param publisher: publisher full name
+        Format: name
```
jacquerie: (What's this?)

inspire_schemas/builders.py
```diff
@@ -226,6 +226,75 @@ def _add_affiliations(author, affiliations):
...
+
+        imprint['date']= date
+        imprint['place']= place
+        imprint['publisher']= publisher
+        self.record.setdefault('imprints', [])
+
+        self.record['imprints'].append(imprint)
+
+    @filter_empty_parameters
+    def add_isbns(self,isbn):
```
jacquerie: Oh, and read the PEP-8: https://www.python.org/dev/peps/pep-0008/

david-caro: I strongly recommend using a plugin for your editor to detect pep8 issues, and pylint.

inspire_schemas/builders.py
```diff
@@ -226,6 +226,75 @@ def _add_affiliations(author, affiliations):
...
+        """
+        :param isbns: the isbns of the book
+        :type isbns: list
+        """
+        isbns={
+            'value': isbn['code'],
+            'medium': isbn['type_of_isbn']
+            }
```
david-caro: maybe this should be a list of dicts?

david-caro: Actually nop, it's appended later to a list :)

inspire_schemas/builders.py
michamos: also, you say `isbns` is a list, but you access it like a dictionary later. If it's the latter, maybe you should say what kind of dictionary is expected? or even flatten the whole thing as

```python
def add_isbn(self, isbn, material=None):
```
? If you do this, make sure to check for emptiness of `material` before adding it to the record.

inspire_schemas/builders.py
```diff
@@ -227,6 +226,79 @@ def _add_affiliations(author, affiliations):
...
+        """
+
+        imprint = {}
+        if date is not None:
+            imprint['date'] = date
+        if place is not None:
+            imprint['place'] = place
+        if publisher is not None:
+            imprint['publisher'] = publisher
+        self.record.setdefault('imprints', [])
```
michamos: you should add `if imprint:` before this in order not to create empty lists, or better yet, define
an `_append_to` method and use it in the following way:
```diff
diff --git a/inspire_schemas/builders.py b/inspire_schemas/builders.py
index 6597419..f2675a8 100644
--- a/inspire_schemas/builders.py
+++ b/inspire_schemas/builders.py
@@ -33,6 +33,8 @@ import idutils
 from .utils import (normalize_author_name_with_comma, normalize_date_iso,
                     validate)
 
+EMPTIES = [None, '', [], {}]
+
 
 def filter_empty_parameters(func):
     """Decorator that is filtering empty parameters.
@@ -42,9 +44,8 @@ def filter_empty_parameters(func):
     """
     @wraps(func)
     def func_wrapper(self, *args, **kwargs):
-        empties = [None, '', [], {}]
         my_kwargs = {key: value for key, value in kwargs.items()
-                     if value not in empties}
+                     if value not in EMPTIES}
 
         if (
             list(my_kwargs.keys()) == ['source'] or not list(my_kwargs.keys())
@@ -120,6 +121,20 @@ class LiteratureBuilder(object):
             return source
         return self.source
 
+    def _append_to(self, field, element):
+        """Append the ``element`` to the ``field`` of the record.
+
+        This method is smart: it does nothing if ``field`` is empty and
+        creates ``field`` if it does not exit yet.
+
+        :param field: the name of the field of the record to append to
+        :type field: string
+        :param element: the element to append
+        """
+        if element not in EMPTIES:
+            self.record.setdefault(field, [])
+            self.record.get(field).append(element)
+
     def validate_record(self):
         """Validate the record in according to the hep schema."""
         validate(self.record, 'hep')
@@ -250,9 +265,7 @@ class LiteratureBuilder(object):
             imprint['place'] = place
         if publisher is not None:
             imprint['publisher'] = publisher
-        self.record.setdefault('imprints', [])
-
-        self.record['imprints'].append(imprint)
+        self._append_to('imprints', imprint)
 
     @filter_empty_parameters
     def add_isbns(self, isbn):
```
I leave it to you to amend the other places where this would be useful and write tests for `_append_to`.

inspire_schemas/builders.py
```diff
@@ -186,7 +201,6 @@ def add_author(self, author):
         produces
         """
         self.record.setdefault('authors', [])
```
michamos: you should use `_append_to` here.

inspire_schemas/builders.py
michamos: you didn't address this.

inspire_schemas/builders.py
michamos: you don't need this line now that you are using `_append_to`.

inspire_schemas/builders.py
```diff
@@ -227,6 +241,78 @@ def _add_affiliations(author, affiliations):
...
+        if place is not None:
+            imprint['place'] = place
+        if publisher is not None:
+            imprint['publisher'] = publisher
+        self.record.setdefault('imprints', [])
+
+        self._append_to('imprints', imprint)
+
+    @filter_empty_parameters
+    def add_isbn(self, isbn):
```
michamos: `isbn` is still an object here instead of flattening it as I suggested earlier.

inspire_schemas/builders.py
```diff
@@ -227,6 +241,78 @@ def _add_affiliations(author, affiliations):
...
+        :param title: the title of the book
+        :type title: string
+        """
+        book_series = {}
+        if title is not None:
+            book_series['title'] = title
+        if volume is not None:
+            book_series['volume'] = volume
+
+        self.record.setdefault('book_series', [])
```
michamos: not needed.

inspire_schemas/builders.py
```diff
@@ -227,6 +241,78 @@ def _add_affiliations(author, affiliations):
...
+        self._append_to('book_series', book_series)
+
+    @filter_empty_parameters
+    def add_book_edition(self, edition):
+        """
+        :param edition: the edition of the book
+        :type edition: string
+        """
+
+        self.record.setdefault('editions', [])
```
michamos: you can use `_append_to` here.

tests/unit/test_hep_builder.py
```diff
@@ -69,3 +70,28 @@
 )
 def test_is_citeable(expected_result, formdata):
     assert is_citeable(formdata) is expected_result
+
+
+@pytest.mark.parametrize(
```
michamos: could you make this a bit more readable?
- two cases are probably enough, one with empty data, one with non-empty data, instead of parameterizing.
- use a different value for the source and for the field to append to.

michamos:
Merged in  367132e

PR #127, opened 2017-03-20
builders.LiteratureBuilder: add pubinfo_freetext

david-caro:
* We were missing the pubinfo_freetext from the publication_info.

### DIFF
--- a/inspire_schemas/builders.py
+++ b/inspire_schemas/builders.py
@@ -270,7 +270,8 @@ class LiteratureBuilder(object):
         page_start=None,
         journal_issue=None,
         journal_title=None,
-        journal_volume=None
+        journal_volume=None,
+        pubinfo_freetext=None,
     ):
         """Add publication info.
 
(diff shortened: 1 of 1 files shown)


michamos:
The freetext is a legacy thing that should not be populated anymore. Why is
this needed?

Il 20 mar 2017 8:43 PM, "David Caro" <notifications@github.com> ha scritto:

david-caro:
it's in the schema, so I thought it was still needed. From the crawler is the only thing we are actually populating right now (in the publication_info entry).
If that's not needed anymore, I can just completely ignore the publication info.

michamos:
Wait, publication _nfo is needed, only the freetext part is not. Cannot hepcrawl build a proper publication_info?

david-caro:
It's done by refextract... that runs at the end in legacy (as the one in labs is supposedly not yet ready)

michamos:
That sounds a bit crazy, but anyhow, this is not a blocker for deploying, right?

david-caro:
Kindof, as the crawling needs this, and the crawling is needed for the deploy

david-caro:
Sorry, it's done in the workflow before sending to legacy if I'm not mistaken

Issue #131, opened 2017-03-24
schema: /acquisition_source/datetime does not have a type


harunurhan:
This part of schema `acquisition_source/datetime` doesn't have `type` which should be `string`
```
"acquisition_source": {
      "$schema": "http://json-schema.org/schema#", 
      "additionalProperties": false, 
      "description": "Only the first source is stored: if the record later gets enriched with\nmetadata coming from a second source, the `acquisition_source` is not\nupdated.\n\n:MARC: ``541``", 
      "properties": {
        "datetime": {
          "description": "This does not necessarily coincide with the creation date of the\nrecord, as there might be some delay between the moment the\noriginal information is obtained and a record is finally created in\nthe system.\n\n:MARC: ``541__d``", 
          "format": "date-time", 
          "title": "Date on which the metadata was obtained"
        },...
```

jacquerie:
Right, we forgot it. We have a small issue before delivering this fix: we need to bump the schemas that `inspire-next` is using to ~31.0, otherwise this fix cannot be applied.

jacquerie:
> we need to bump the schemas that `inspire-next` is using to ~31.0

This is happening in https://github.com/inspirehep/inspire-next/pull/2134.

PR #132, opened 2017-03-24
elements: add missing types

jacquerie:
Closes https://github.com/inspirehep/inspire-schemas/issues/131

### DIFF
--- a/tests/integration/fixtures/hep_example.json
+++ b/tests/integration/fixtures/hep_example.json
@@ -132,11 +132,7 @@
         }
     ],
     "acquisition_source": {
-        "datetime": {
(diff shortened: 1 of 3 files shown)


tests/integration/fixtures/hep_example.json
```diff
@@ -133,9 +133,7 @@
     ],
     "acquisition_source": {
         "datetime": {
-            "description": "This does not necessarily coincide with the creation date of the\nrecord, as there might be some delay between the moment the\noriginal information is obtained and a record is finally created in\nthe system.\n\n:MARC: ``541__d``",
-            "format": "date-time",
-            "title": "Date on which the metadata was obtained"
```
jacquerie: Note here that not specifying the `type` causes some very weird behavior...

jacquerie: ...but actually this is a good thing, because now we can grep for `description` in `tests/integration/fixtures` to find all the other places in which we missed a type.

michamos:
:scream: this was an incompatible change!

PR #134, opened 2017-03-29
Builder edited to add book information

Glignos:
added the needed functions to the builder to add book information


michamos:
duplicate of #124

PR #140, opened 2017-04-12
Related records

michamos:
* INCOMPATIBLE Adds `related_records` for hep, experiments, journals and
institutions instead of the previous special cases (`succeeding_entry`,
`related_experiments`, `relation`, `related_institutes` respectively)
and the corresponding `related_record` element.

This is what we discussed with @annetteholtkamp and @jacquerie.

### DIFF
--- a/inspire_schemas/records/elements/related_record.yml
+++ b/inspire_schemas/records/elements/related_record.yml
@@ -1,0 +1,93 @@
+$schema: http://json-schema.org/schema#
+additionalProperties: false
+dependencies:
+    record:
+        oneOf:
+        -   required:
+            - relation
+        -   required:
+            - relation_freetext
+description: |-
+    :MARC: ``78002``, ``78502``, ``78708```and ``510`` (depending on type of
+        record and :ref:`related_record.json#/properties/relation`)
+properties:
+    curated_relation:
+        type: boolean
+    record:
+        $ref: json_reference.json
+        description: |-
+            :MARC: ``510__0`` (for Institutions and Experiments),
+                ``78002/78502/78708w`` (for Literature and Journals).
+        title: URL of related record
+    relation:
+        description: |-
+            The possible values are:
+
+            ``predecessor``
+                :MARC: ``510__w:a`` (if not a Literature record) or the field
+                    comes from ``78002`` (for Literature).
+
+                The related record is a predecessor of the current one, i.e.
+                the current record supersedes the related one.
+
+                - For a Literature record, this is typically the case if the
+                  related record is a preliminary note and this record is the
+                  final version.
+                - For a Journal record, this happens when the journal has
+                  changed name: the related record has the previous name and
+                  this one the new name.
+                - For an Institution record, this means that the related record
+                  represents an institution that does no longer exist and has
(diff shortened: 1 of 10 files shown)


inspire_schemas/records/elements/related_record.yml
```diff
@@ -0,0 +1,89 @@
+$schema: http://json-schema.org/schema#
+additionalProperties: false
+dependencies:
+    record:
+        oneOf:
+        -   required:
+            - relation
+        -   required:
+            - other_relation
```
kaplun: Woah! Havey you checked with @harunurhan? Will it break current editor?

inspire_schemas/records/elements/related_record.yml
```diff
@@ -0,0 +1,89 @@
...
+description: |-
+    :MARC: ``78002``, ``78502``, ``78708```and ``510`` (depending on type of
+        record and :ref:`related_record.json#/properties/relation`)
+properties:
+    curated_relation:
+        type: boolean
+    other_relation:
```
kaplun: Reading `other_relation` I am not understanding this has to be used if `relation` does not fit.

What about `other_relation_type` or `relation_other`?

tests/integration/fixtures/experiments_example.json
```diff
@@ -1,122 +1,139 @@
...
     "core": true,
-    "date_approved": "dddd-dd-dd",
-    "date_cancelled": "dddd-dd-dd",
-    "date_completed": "dddd-dd-dd",
-    "date_proposed": "dddd-dd-dd",
-    "date_started": "dddd-dd-dd",
-    "deleted": true,
+    "date_approved": "9433-07-69",
+    "date_cancelled": "0653-71-44",
+    "date_completed": "5802-68-29",
```
kaplun: These dates are bit ugly... would be great to have real valid ones... Can be fixed in a different issue.

inspire_schemas/records/elements/related_record.yml
michamos: it was done this way on purpose not to break the editor.
Morally speaking, we want to have a big
```yaml
allOf:
    - oneOf:
        - properties:
            relation:
                # definition of relation
        - properties:
            other_relation:
                # definition of other_relation
    - properties:
        record:
            # definition of record
```
but **that** breaks the record editor.
Instead, I used a hack by putting dependencies on `record` which is a required field. To be valid, `record` should be present, which in turns triggers its dependencies. So the effect is the same, but the editor does not have to deal with it before validation.

tests/integration/fixtures/experiments_example.json
michamos: before it was producing `dddd-dd-dd` always, so this is already better I would say.

inspire_schemas/records/elements/related_record.yml
michamos: It is not clear to me whether inverting the two words or adding `type` caries a much clearer meaning, but if you think so, I can change it. Maybe `relation_type`/`other_relation_type` for consistency then?

michamos:
@kaplun I addressed your comments.

PR #141, opened 2017-04-18
conferences: finalize schema

michamos:
Generated docs from the schema are here https://michamos.github.io/inspire-schemas/branch/conference_schema/records/conferences.html.

### DIFF
--- a/inspire_schemas/records/conferences.yml
+++ b/inspire_schemas/records/conferences.yml
@@ -5,52 +5,119 @@ properties:
         format: url
         type: string
     _collections:
+        description: |-
+            Used by `invenio-collections` to store the collections this record
+            belongs to.
+
+            .. note::
+
+                This field is maintained by `invenio-collections` and should
+                not be edited manually.
         items:
             type: string
+        title: '`invenio-collections` metadata'
         type: array
     _private_notes:
+        description: |-
+            :MARC: ``595``
+
+            These notes are only visible to privileged users, not regular
+            users.
         items:
             $ref: elements/sourced_value.json
+        title: List of private notes
         type: array
         uniqueItems: true
-    acronym:
+    acronyms:
         items:
+            description: |-
+                :MARC: ``111__e``
+                :example: ``SUSY 2018``
+            title: Conference acronym
             type: string
-        title: Conference acronym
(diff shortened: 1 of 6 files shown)


inspire_schemas/records/elements/address.yml
```diff
@@ -18,13 +32,37 @@ properties:
...
+
+              :MARC: Not present.
+              :example: ``U. Libre Brussels``
         type: string
+    postal_address:
+        description: |-
+            Full postal address in the original language, with
+            one element for every line, but excluding the country.
+
+            :example: ``["Notkestraße 85", "22607 Hamburg"]``
```
kaplun: I think we shall also exclude the ZIP code, non?

michamos: this is done on purpose, because the formatting of the postal code + city changes between countries.
The idea is that one can take this and use it to write a letter to the institution (don't know how useful that is in practice, though).

inspire_schemas/records/conferences.yml
```diff
@@ -5,52 +5,110 @@ properties:
...
         items:
             $ref: elements/address.json
+        title: List of locations of the conference
         type: array
         uniqueItems: true
     alternative_titles:
+        description: |-
+            :MARC: ``711``
+
+            Sometimes conferences are known by several names/acronyms. These
```
michamos: yeah, also titles could go to titles. Currently, the `711` field on legacy contains lots of crap that are used to boost search results (roman numbers, variants of acronyms with/without spaces between the name and the number, etc.) that are put into this field. Maybe the name of this field could be made more clear?

kaplun: `crap`?

michamos: things that are not necessarily titles or acronyms, but some pieces of them, http://inspirehep.net/search?ln=en&cc=Conferences&ln=en&cc=Conferences&p=711%3A**&action_search=Search&sf=conferencestartdate&so=d&rm=&rg=250&sc=0&of=hb&ot=711.

kaplun: Was just joking, suggesting to call the field `crap` :dancing_women:

PR #145, opened 2017-04-21
hep: use enum for energy ranges

michamos:
Based on @ksachs's input.

### DIFF
--- a/inspire_schemas/records/hep.yml
+++ b/inspire_schemas/records/hep.yml
@@ -684,12 +684,26 @@ properties:
         uniqueItems: true
     energy_ranges:
         description: |-
-            :MARC: ``695__e``
+            :MARC: ``695__e:n`` corresponds to the ``n``-th range in the enum,
+                e.g. ``695__e:2`` is `3-10 Gev``.
 
-            Ranges of energies the record refers to.
+            Ranges of energies the record refers to. The ranges are inclusive
+            on the right.
+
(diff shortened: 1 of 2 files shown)


inspire_schemas/records/hep.yml
```diff
@@ -684,12 +684,21 @@ properties:
...
-            type: integer
+            enum:
+            - 0-3 GeV
+            - 3-10 GeV
+            - 10-30 GeV
+            - 30-100 GeV
+            - 100-300 GeV
+            - 300-1000 GeV
+            - 1-10 TeV
+            - '> 10 TeV'
```
michamos: this is quoted because `>` has special meaning in yaml.

david-caro: \o/ hurray yaml extensive set of hidden/quirk features!

david-caro: Are the ranges inclusive to the left?
ex. 3-10 == (>=3 && <10) ?

david-caro: Maybe the comment should say something about that (I would be confused, I'm not a curator though...

ksachs: it's inclusive on the right. I.e. 695__e:6 ( 300-1000 GeV) includes 1 TeV

michamos: I documented it in an amended commit. Could someone approve this?

PR #153, opened 2017-05-15
schemas: promote acquisition_source to array

michamos:
* INCOMPATIBLE The field `acquisition_source` is replaced by the
`acquisition_sources` array, allowing to capture all sources of metadata
in the record instead of only one. This is useful as a record typically
aggregates information from several sources.

### DIFF
--- a/inspire_schemas/builders.py
+++ b/inspire_schemas/builders.py
@@ -108,7 +108,7 @@ class LiteratureBuilder(object):
     def _append_to(self, field, element):
         """Append the ``element`` to the ``field`` of the record.
 
-        This method is smart: it does nothing if ``field`` is empty and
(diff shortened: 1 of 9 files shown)


inspire_schemas/builders.py
```diff
@@ -118,6 +118,20 @@ def _append_to(self, field, element):
             self.record.setdefault(field, [])
             self.record.get(field).append(element)
 
+    def _prepend_to(self, field, element):
+        """Prepend the ``element`` to the ``field`` of the record.
+
+        This method is smart: it does nothing if ``field`` is empty and
+        creates ``field`` if it does not exit yet.
```
kaplun: You mean if `element` is emtpy.

michamos:
@kaplun I have just changed the wrong docstrings.

PR #154, opened 2017-05-16
Institution schema

michamos:
Schema is done and documented, but documentation generation failed because of some `jsonschema2rst` bug, so you will have to read the YAML.

### DIFF
--- a/inspire_schemas/records/institutions.yml
+++ b/inspire_schemas/records/institutions.yml
@@ -6,68 +6,111 @@ properties:
         type: string
     ICN:
         description: |-
-            HEP affiliation following new standards
+            :MARC: ``110__t``
+
+            HEP affiliation unique identifier following the general pattern
+            ``institution name, city, department``. This was called the "new ICN"
+            on legacy.
+
+            .. note::
+
+                This is a list in case the ICN needs to be updated because of a
+                mistake in its assignment.  The current one is the first
+                element of the list.
         items:
             type: string
-        title: New INSPIRE control number
+        title: List of affiliation identifiers
         type: array
     _collections:
+        description: |-
+            Used by `invenio-collections` to store the collections this record
+            belongs to.
+
+            .. note::
+
+                This field is maintained by `invenio-collections` and should
+                not be edited manually.
         items:
             type: string
+        title: '`invenio-collections` metadata'
         type: array
     _private_notes:
+        description: |-
+            :MARC: ``595``, ``667``
+
+            These notes are only visible to privileged users, not regular
+            users.
         items:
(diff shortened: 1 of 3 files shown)


inspire_schemas/records/institutions.yml
```diff
@@ -6,76 +6,122 @@ properties:
...
                         type: string
                     value:
+                        description: |-
+                            :MARC: ``035__a``
                         pattern: ^\d+$
                         type: string
                 required:
                 - schema
                 - value
+                title: HAL identifier
```
kaplun: Is it correct that this is indented at this level while the one for SPIRES is not?

inspire_schemas/records/institutions.yml
```diff
@@ -85,75 +131,132 @@ properties:
...
+            properties:
+                acronym:
+                    description: |-
+                        :MARC: in parenthesis at the end of ``110_a/b``
+                    type: string
+                name:
+                    description: |-
+                        :MARC: value in ``110_a/b``, excluding final parenthesis
+                    type: string
+            type: object
```
kaplun: Don't we also link in line to the corresponding parent?

inspire_schemas/records/institutions.yml
```diff
@@ -166,23 +269,13 @@ properties:
         uniqueItems: true
     self:
         $ref: elements/json_reference.json
-        description: |-
-            Url of the record itself
-        title: Url of the record
-    time_zone:
```
kaplun: No? We have killed it?

inspire_schemas/records/institutions.yml
michamos: no, I will change the SPIRES one

inspire_schemas/records/institutions.yml
michamos: we link using `related_records`, do you think it is better to put the link here?

inspire_schemas/records/institutions.yml
michamos: This was used to time some e-mail campaigns. If it is ever useful again, we can deduce it from the address of the institution.

inspire_schemas/records/institutions.yml
kaplun: I think it's necessary, in particular if we want to update these piece of information automatically from the related records.

michamos: after discussing IRL, we cannot do this as it would change the format from the used on legacy in a way that cannot be solved by doJSON.

michamos:
I addressed all your comments.

PR #160, opened 2017-05-24
builder: support `dois.material` field

### LINKED ISSUE #159: builder: support `dois.material` field
Builder isn't populating `dois.material` field.

spirosdelviniotis:
* Adds `dois.material` field to tests.
* Adds: support for `dois.material` field in `LiteratureBuilder`.

Closes #159
Relates to https://github.com/inspirehep/hepcrawl/issues/128

Signed-off-by: Spiros Delviniotis <spirosdelviniotis@gmail.com>

### DIFF
--- a/inspire_schemas/builders.py
+++ b/inspire_schemas/builders.py
@@ -170,7 +170,7 @@ class LiteratureBuilder(object):
         self.set_citeable(True)
 
     @filter_empty_parameters
-    def add_doi(self, doi, source=None):
+    def add_doi(self, doi, source=None, material=None):
         """Add doi.
 
         :param doi: doi for the current document.
@@ -178,12 +178,21 @@ class LiteratureBuilder(object):
 
         :param source: source for the doi.
         :type source: string
+
+        :param material: material for the doi.
+        :type material: string
         """
-        if idutils.normalize_doi(doi):
-            self._append_to('dois', {
-                'value': doi,
-                'source': self._get_source(source),
-            })
+        if not idutils.normalize_doi(doi):
+            return
+
+        dois = {
+            'value': doi,
+            'source': self._get_source(source)
+        }
+        if material is not None:
+            dois['material'] = material
+
+        self._append_to('dois', dois)
 
     @filter_empty_parameters
     def add_author(self, author):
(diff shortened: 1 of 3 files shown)


inspire_schemas/builders.py
```diff
@@ -170,17 +170,21 @@ def add_arxiv_eprint(self, arxiv_id, arxiv_categories):
...
 
         :param source: source for the doi.
         :type source: string
+
+        :param material: material for the doi.
+        :type material: string
         """
         if idutils.normalize_doi(doi):
             self._append_to('dois', {
+                'material': material,
```
david-caro: I don't think you should put the material if None is passed, it should be skipped then instead.

david-caro: See https://github.com/inspirehep/inspire-schemas/pull/160/commits/a001ae48a56a406c6c3ffc14fb76857ba53467fc#diff-4e3333cd3c68d78af7fef5328208f014L236 for example

spirosdelviniotis: You are right!
What is the example? I don't see it..

david-caro: You have to expand the diff.... github is sometimes messy.
It's the function `def add_book(self, publisher=None, place=None, date=None):`

inspire_schemas/builders.py
```diff
@@ -170,20 +170,26 @@ def add_arxiv_eprint(self, arxiv_id, arxiv_categories):
...
         :type doi: string
 
         :param source: source for the doi.
         :type source: string
+
+        :param material: material for the doi.
+        :type material: string
         """
+        dois = {}
         if idutils.normalize_doi(doi):
```
david-caro: Id negate the if here, to avoid having all the code indented, like:
```python
if not idutils.normalize_doi(doi):
    return

dois = {}
...
```

david-caro: It's relevant here to remember that:
* That is ok if most of the code is inside the if
* You should put the eager return as soon as possible in the code, so it's clear that it will not run on certain conditions (if it's not in the beginning of the code, the if is easier to miss).

If one of those can't be fulfilled, then it's better to leave it as it was ;)

spirosdelviniotis: Thank you very much! nice comment @david-caro :smile:

PR #162, opened 2017-05-24
Add parent_record field to builder

Glignos:
* Add parent record field to builder (publication_info) support #2159


jmartinm:
LGTM

PR #169, opened 2017-06-12
experiments: make experiment.value not required

### LINKED ISSUE #168: experiments: experiment.value can't be required
Currently `experiment.value` is required: https://github.com/inspirehep/inspire-schemas/blob/006537d31a1267959ac7003805aecdb04359434a/inspire_schemas/records/experiments.yml#L203-L204.

It shouldn't be, because no record has the `119__c` field that should populate it: https://inspirehep.net/search?cc=Experiments&p=119__c%3A**&wl=0.

jacquerie:
Closes #168


jacquerie:
NB: to be merged after #170

PR #172, opened 2017-06-13
hep: enlarge some enums

michamos:
* Adds the CDS hidden collection and BABAR internal collections to
`special_collections` and clarifies its description.
* Adds the INIS keyword schema.

### DIFF
--- a/inspire_schemas/records/hep.yml
+++ b/inspire_schemas/records/hep.yml
@@ -852,6 +852,13 @@ properties:
                         Describes to which vocabulary the keyword in :ref:`value`
                         belongs.
 
+                        ``INIS``
+                            :MARC: ``695__2:INIS``
+
+                            The keyword is part of the `International
+                            Nuclear Information System (INIS)
+                            <https://www.iaea.org/inis/>`_ vocabulary.
+
                         ``INSPIRE``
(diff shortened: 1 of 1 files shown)


inspire_schemas/records/hep.yml
```diff
@@ -1287,13 +1295,18 @@ properties:
         description: |-
             :MARC: ``980__a``
 
-            These collections are used by collaborations to manage their notes.
+            These collections are used by different groups to manage records
+            using the Inspire infrastructure, but which do not belong to
+            the Inspire litterature collection.
```
jacquerie: `s/litterature/literature/`

inspire_schemas/records/hep.yml
```diff
@@ -1287,13 +1295,18 @@ properties:
...
             If a record belongs to a special collection, it is not shown in the
             standard search results.
         items:
             enum:
+            - BABAR-ANALYSISDOCUMENT
```
jacquerie: What about `BABAR-ANALYSIS-DOCUMENT` ?

michamos: that's a special case in the conversion, but if you are fine with that I can change it.

PR #173, opened 2017-06-13
elements: new language_code enum

michamos:
* INCOMPATIBLE Adds a language_code element to restrict the allowed
languages better than with a pattern. Its values are ISO 639-1 language
codes, generated by filtering the list of `pycountry.languages` for
those having an alpha_2 code.


inspire_schemas/records/elements/language_code.yml
```diff
@@ -0,0 +1,195 @@
...
+- na
+- nv
+- nr
+- nd
+- ng
+- ne
+- nl
+- nn
+- nb
+- 'no'
```
michamos: YAML quirk again, as unquoted `no` corresponds to JSON `false`.

kaplun:
Is it really INCOMPATIBLE? There is not much to do `inspire-next` wise, beside cleaning up failing metadata.

michamos:
Well technically it is: a record with an integer-valued energy range is
not valid with this new schema. I agree that it is a very mild
incompatibility though.

kaplun:
Wait I think you are replying to another ticket? This one is about language enum.

michamos:
Whoops, I shouldn't reply to tickets this late in the evening :)

Anyway, here it is also technically incompatible: any 2-character
string was valid, whereas now it needs to be part of the enum. In
practice it shouldn't change anything, as we should already be
producing only valid languages.

kaplun:
Yeah, exactly. So for me: LGTM. So this is just a headsup for @jacquerie that, in case you maintain the INCOMPATIBLE tag when this is merged, is not really anything worth thinking about to bump the major version on inspire-next.

jacquerie:
INCOMPATIBLE has an objective definition, which this commit satisfies, so it must stay INCOMPATIBLE. It doesn't matter if bumping the schema is as trivial as modifying `setup.py`, it must still be done...

kaplun:
I am not sure we ever defined explicitly what INCOMPATIBLE is for inspire-schemas. So we agree that everytime the validity of the schema is altered or made stricter the change is going to be declared INCOMPATIBLE even if no client source code is affected?

jacquerie:
> I am not sure we ever defined explicitly what INCOMPATIBLE is for inspire-schemas.

A commit is INCOMPATIBLE when merging it might cause an existing record that satisfies the previous schema to become invalid.

PR #176, opened 2017-06-16
Introducing schema revisions

### LINKED ISSUE #9: Schema versioning
Schemas will evolve with time. For this reason we should introduce versioning. Each schema should be versioned with `semver` technique. Each time a modification is performed this is done by copying the last schema into a new file first and then performing the modification.

### LINKED ISSUE #10: Record migrators
After versioning is introduced in #9 we will need to support upgrading records. There should be an API similar to:
```python
def record_needs_upgrade(record):
     ...
     return True

def upgrade_record(record):
     ...
     return upgraded_record
```

kaplun:
Closes #9, Closes #10 


kaplun:
@jacquerie @michamos @david-caro this is an initial proposal (still implementing) for supporting ledgers and revisions

kaplun:
@jacquerie @michamos @david-caro this is an initial proposal (still implementing) for supporting ledgers and revisions. Note this is super WIP

michamos:
new element added in #173

tests/unit/test_utils.py
```diff
@@ -155,3 +155,29 @@ def test_load_schema_without_schema_key(mock_get_schema_path, mock_open):
...
+def test_build_latest_schema_revisions():
+    latest_schema_revisions = utils.build_latest_schema_revisions()
+
+    assert 'literature' in latest_schema_revisions
+
+    # Testing journals because the max revision will not change that often
+    assert latest_schema_revisions['journals'] == '0.0.1'
+
+
+def test_get_schema_and_revision():
```
david-caro: This looks like a good candidate to use `pytest.mark.parametrize` thingie.

inspire_schemas/upgraders.py
```diff
@@ -0,0 +1,85 @@
...
+        'journals': [
+            ('0.0.1', first_revision_rule)
+        ],
+    }
+
+
+LEDGERS = build_ledger()
+
+
+def upgrade(json):
```
michamos: what is 'json'? calling `rule(json)` suggests it is a record, but `get_schema_and_revision(json)` suggests that it is a string instead.

inspire_schemas/upgraders.py
```diff
@@ -0,0 +1,85 @@
...
+    schema, current_revision = get_schema_and_revision(json)
+    for revision, rule in LEDGERS['schema']:
```
michamos: This only works if the versions in the ledger are ordered from smallest to largest. This should probably be tested and either documented or sorted in the right order in this loop.

inspire_schemas/upgraders.py
```diff
@@ -0,0 +1,85 @@
...
+        if semver.compare(revision, current_revision, loose=False) > 0:
+            json = rule(json)
+    json['schema'] = '{schema}-{revision}.json'.format(
+        schema=schema,
+        revision=LATEST_SCHEMA_REVISIONS[schema]
+    )
+    return json
+
+
+def first_revision_rule(json):
```
michamos: alternatively, we could use descriptive names like: `new_foo_field`.

tests/unit/test_utils.py
michamos: you could also simplify to
```python
result = utils.get_schema_and_revision('literature-1.2.3.json')
expected = 'literature', '1.2.3'
assert result == expected
```
instead of unpacking the tuples.

michamos:
BTW, is semver really needed here? if the change is not incompatible, the records don't need to be upgraded.

kaplun:
@david-caro  I think this branch can be considered dead? Has it been decided how to move on with versioning?

david-caro:
Well, we decided long ago to go with this, but nothing has been done so far (that I know of)

michamos:
I think @jmartinm wanted to investigate whether we could use alembic instead for that, but don't know if he ever did.

jacquerie:
We can close this.

PR #182, opened 2017-06-22
Author schema changes

michamos:
I worked on this some time ago with @annetteholtkamp, but I never pushed because it is not strictly speaking final as we want to have user feedback about what there should be in author profile pages.
It still contains some useful refactorings, with useless fields removed and a clearer organization, so it might serve as a starting point.

This should probably not be merged before gathering the requirements as it will require changes in doJSON and in the formdata to model, which we might not want to do twice.

### DIFF
--- a/inspire_schemas/records/authors.yml
+++ b/inspire_schemas/records/authors.yml
@@ -131,9 +131,45 @@ properties:
     name:
         additionalProperties: false
         description: |-
-            Contains name information.
+            :MARC: ``100``, ``400``, ``880``
         properties:
+            name_variants:
+                description: |-
+                    Contains other variations (besides `preferred_name`) of the
+                    author name in `value` that are in use.
+
+                    These could be:
+
+                    - other spellings;
+                    - other transliterations from the :ref:`native_names` to the
+                      Latin alphabet;
+                    - other splittings among last names and first names;
+                    - combinations with :ref:`previous_names`.
+
+                    :example: ``Smith-Taylor, Johnny``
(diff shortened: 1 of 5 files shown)


jacquerie:
IMO after https://github.com/inspirehep/inspire-dojson/pull/66 is

- [ ] reviewed
- [ ] merged
- [ ] integrated in `inspire-next`
- [ ] tested on all Journals

this should go in: it's better to have a semi-stable schema than not.

inspire_schemas/records/authors.yml
```diff
@@ -155,70 +262,71 @@ properties:
...
-            type: string
-        minItems: 1
-        type: array
-        uniqueItems: true
+            :MARC: ``970__d``
+
+            Contains a reference to the record replacing the current one, if it
+            is marked as :ref:`deleted`.
+        title: Record replacing this one
     positions:
```
drjova: flatten the institution history object

michamos: done, it's not exactly the same though (`name` vs `institution`).

scripts/generate_example_records.js
```diff
@@ -26,7 +26,6 @@ jsf.format('orcid', function(gen, schema){
     }
     return almost_orcid + ck
 });
-jsf.format('.+, .+', function(gen, schema){ return gen.randexp('^.+, .+$');});
```
drjova: I didn't know we have a js for this :P

david-caro:
Merging and bumping the major version so we can start working on top of it.

PR #187, opened 2017-06-27
experiments: make room for legacy accelerators

michamos:
* INCOMPATIBLE Moves `legacy_name` to top-level in order to take into
account that not only experiments, but also accelerators can have a
`legacy_name`.
* Also documents how to spot an `accelerator`.


jacquerie:
> how to spot an `accelerator`

They're usually quite hard to miss...

Issue #190, opened 2017-06-29
travis: the Python 3 build is flaky


jacquerie:
The Python 3 build fails occasionally with the following error: https://travis-ci.org/inspirehep/inspire-schemas/jobs/248543416#L416

michamos:
This is not happening anymore, now it's consistently failing because of doctests with unicode literals.

jacquerie:
Right, this issue is obsolete.

PR #191, opened 2017-06-29
travis: allow failures on Python 3

jacquerie:
Addresses https://github.com/inspirehep/inspire-schemas/issues/190

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ notifications:
 sudo: false
 language: python
 matrix:
+  allow_failures:
+    - python: '3.6'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,6 +3,8 @@ notifications:
 sudo: false
 language: python
 matrix:
+  allow_failures:
+    - python: '3.5'
```
michamos: Maybe update to `python 3.6` while you're at it?

jacquerie: Sure, just added a commit for that.

PR #195, opened 2017-07-06
material: add corrigendum

kaplun:

### DIFF
--- a/inspire_schemas/records/elements/material.yml
+++ b/inspire_schemas/records/elements/material.yml
@@ -6,5 +6,6 @@ enum:
 - erratum
 - addendum
 - translation
+- corrigendum
 title: Material to which the field refers
 type: string


kaplun:
Fixes: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/822798/events/150768065/

michamos:
Do you really have to do this while Annette and I are away? This is the
wrong solution, we don't want to distinguish errata and corrigenda. So the
proper fix is to have dojson translate `corrigendum` to `erratum` instead.

Il 06 lug 2017 12:52 PM, "Samuele Kaplun" <notifications@github.com> ha
scritto:

kaplun:
I was deputizing you, since I keep on seeing the Sentry errors.
In this case I verified that the original source (Hindawi in this case) is calling this `corrigendum`.
But indeed if you want to squeeze `corrigendum` into `erratum` then I will revert this and we need to solve it into `inspire-dojson`.

Issue #200, opened 2017-07-18
Add files entries in schemas


StellaCh:
The plots are generated by plotextract, and need the following metadata:

- format: png, jpeg, ...
- label: The name that was given to it in the pdf
- caption: A 'join' of all the captions that image has (normally only one).
- name: The name of the plot
- key:The 'key' in the record.files "property"
- material: form the enum
- source: common source element, with the info of the source of the file.

The documents will have the following metadata:

- hidden: if the document should be hidden to the general public (true|false)
- material: the usual material field, if it's preprint, etc
- url: original url of the document if any
- source: common source element, with the info of the source of the file.
- key:The 'key' in the record.files "property"
- description: what’s inside the document
- fulltext: flag, whether it's the full content of the record, as opposed to, say, a table of contents

michamos:
@david-caro @zzacharo @ammirate  do we really want to store the format of the plot there? isn't this computed already by `invenio-files-rest`?

michamos:
And what is the `name` of the plot for?

StellaCh:
As far as I understand, the name was given by the user who uploaded the plot and the key is generated from the system, right @david-caro ?

david-caro:
Yep, the `name` is the name the user gave to it, like `fig1.png`, while the `key` is what we use to uniquely identify that file (as two plots might have the same `name`, but not the same `key`).

> @david-caro @zzacharo @ammirate do we really want to store the format of the plot there? isn't this computed already by invenio-files-rest?

It's in the schema, but afaik it does not (https://labs.inspirehep.net/api/holdingpen/682187)... though I might be missing something. In any case, the idea of having the metadata we need/want for the plots and documents in a separated key allows us to decouple it from invenio-files-rest (or any other external module for that matter) and evolve it on our own, using the `.files` property as a storage backend only.

PR #201, opened 2017-07-18
hep: handle files properly

michamos:
I didn't add the keys I asked about in #200 (`figures/name` and `figures/type`), if they are needed I will do an update.

### DIFF
--- a/inspire_schemas/records/hep.yml
+++ b/inspire_schemas/records/hep.yml
@@ -106,28 +106,7 @@ properties:
                 This field is maintained by `invenio-records-files` and should
                 not be edited manually.
         items:
-            description: |-
-                Describe information needed for files in records.
-            properties:
-                bucket:
-                    type: string
-                checksum:
-                    type: string
-                key:
-                    type: string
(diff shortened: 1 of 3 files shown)


david-caro:
Yep, they are needed.

michamos:
@david-caro I updated the PR taking into account our discussion IRL. Could you have a look?

inspire_schemas/records/hep.yml
```diff
@@ -650,6 +629,56 @@ properties:
...
+                    description: |-
+                        :MARC: ``8564_y``
+
+                        Describes how this document relates to the document that
+                        this record represents.
+                        :example: ``table of contents``
+                    type: string
+                fulltext:
+                    title: Whether this document contains the fulltext
+                    type: boolean
```
david-caro: Do we want to put anything that's not the fulltext here?

michamos:
yes, parts of the fulltext, e.g. only a table of contents, or related material that is not the fulltext (poster presented at a conference attached to the conference paper).

Issue #142, opened 2017-04-21
Unit tests for the builder


rikirenz:
We do not have the unit tests for the builder. 
This is the list of the unit test that we should write:

- [ ] add_abstract
- [ ] add_arxiv_eprint
- [ ] add_doi
- [ ] add_author
- [ ] make_author
- [ ] add_book
- [ ] add_isbn
- [ ] add_book_series
- [ ] add_book_edition
- [ ] add_inspire_categories
- [ ] add_private_note
- [ ] add_publication_info
- [ ] add_imprint_date
- [ ] add_preprint_date
- [ ] add_thesis
- [ ] add_accelerator_experiments_legacy_name
- [ ] add_language
- [ ] add_license
- [ ] add_public_note
- [ ] add_title
- [ ] add_title_translation
- [ ] add_url
- [ ] add_report_number
- [ ] add_collaboration
- [ ] add_acquisition_source
- [ ] add_document_type
- [ ] add_copyright
- [ ] add_number_of_pages
- [ ] add_special_collection
- [ ] add_publication_type
- [ ] set_core
- [ ] set_refereed
- [ ] set_withdrawn
- [ ] set_citeable

PR #202, opened 2017-07-25
builder: add methods for figures and documents

zzacharo:
Signed-off-by: Zacharias Zacharodimos <zacharias.zacharodimos@cern.ch>


jacquerie:
Looks ok, but we should start using Google-style docstrings and writing tests (https://github.com/inspirehep/inspire-schemas/issues/142).

zzacharo:
I will add the tests for the new methods.

jacquerie:
> I will add the tests for the new methods.

Don't worry, for now it's ok without, so you can continue doing what needs these methods.

PR #205, opened 2017-08-01
Integrate reference builder

michamos:


inspire_schemas/builders/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE-SCHEMAS; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
```
jacquerie: Minor thing, but we might want to fix licenses on this repo to be GPLv3 and to say that is part of INSPIRE, non INSPIRE-SCHEMAS... absolutely not something to be fixed in this PR.

PR #213, opened 2017-08-11
Fixed initials parsing

iulianav:
When importing an article from CrossRef, if the author's name contains initials
each followed by a space, they would be displayed as one word, appendend
together.

The requirement I enforce ensures that a dot is appended after each initial for
readability.

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -31,7 +31,7 @@ from functools import wraps
 
 import idutils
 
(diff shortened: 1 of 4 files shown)


jacquerie:
I probably said too many scary things about `git push --force` : ) 

It's a dangerous command, but it's the right way to update a PR instead of creating a new one. No big problem here, but due to how GitHub handles things, opening multiple PRs for the same feature might lead to lost review comments.

PR #214, opened 2017-08-16
utils: improve author name normalizer

iulianav:
Renamed the function to something more representative, improved the implementation and logic behind the code and reduced vertical space of tests.

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -30,6 +30,7 @@ import re
 import six
 from jsonschema import validate as jsonschema_validate
 from jsonschema import RefResolver, draft4_format_checker
+from nameparser import HumanName
 from pkg_resources import resource_filename
 from unidecode import unidecode
 
@@ -212,7 +213,7 @@ def validate(data, schema=None):
     )
 
 
-def normalize_author_name_with_comma(author):
+def normalize_author_name(author):
     """Normalize author name.
 
     :param author: author name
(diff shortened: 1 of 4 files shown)


tests/unit/test_utils.py
```diff
@@ -191,3 +191,46 @@ def test_split_pubnote():
...
+    assert expected == utils.normalize_author_name('Smith, J.P. ')
+    assert expected == utils.normalize_author_name('Smith, J. P. ')
+
+
+def test_normalize_author_name_with_spaces():
+
+    expected = 'Smith, J.P.'
+
+    assert expected == utils.normalize_author_name('Smith, J P ')
+    assert expected == utils.normalize_author_name('Smith, J P')
```
michamos: can you also add the following tests? don't know whether they will pass or `HumanName` is not smart enough.

```
def test_normalize_author_name_handles_several_last_names():

    expected = 'Smith Davis, J.P.'

    assert expected == utils.normalize_author_name('Smith Davis, J.P.')


def test_normalize_author_name_handles_jimmy():

    expected = 'Jimmy'

    assert expected == utils.normalize_author_name('Jimmy')
```

in case you are wondering: http://people.physics.tamu.edu/jimmy/

inspire_schemas/utils.py
```diff
@@ -208,20 +209,47 @@ def validate(data, schema=None):
...
+    if _is_initial(name.first) and _is_initial(name.middle):
+        normalized_names = '{first_name}{middle_name}'
+    else:
+        normalized_names = '{first_name} {middle_name}'
+
+    normalized_names = normalized_names.format(
+        first_name=name.first, middle_name=name.middle
+        )
+
+    if name.last and (name.first or name.middle):
```
michamos: If I understand correctly, the whole logic here amounts to the fact that if there is both a last name and first names, you should join them with a comma, otherwise take what there is. Why not write that then? something like

```python
normalized_names = normalized_names.strip()
final_name =  [name for name in (normalized_names, name.last) if name].join(', ').strip()

return final_name
```

This uses the fact that [empty strings are falsy in Python](https://docs.python.org/2/library/stdtypes.html#truth-value-testing).

inspire_schemas/utils.py
```diff
@@ -208,20 +209,47 @@ def validate(data, schema=None):
...
+            author_name = ''.join(seq)
+        return author_name
 
-    return name
+    name = HumanName(author)
+
+    name.first = _ensure_dotted_initials(name.first)
+    name.middle = _ensure_dotted_initials(name.middle)
+
+    if _is_initial(name.first) and _is_initial(name.middle):
```
michamos: why the `and ...`?

iulianav: I noticed that in the given example in the issue, whenever both first and middle names are initials, there was no space in between them e.g. 'Smith, J.A.' and NOT 'Smith, J. A.'. However, I am not sure if that is a requirement. Is it @michamos ?

michamos: Yes, that's correct indeed :+1:

iulianav: Great!

jacquerie:
This build instead generated `46.0.2` normally: https://travis-ci.org/inspirehep/inspire-schemas/jobs/267146439#L961, as it should have.

PR #216, opened 2017-08-18
journals: attempt to finalize schema

michamos:
This is a draft for the new journal schema. Any comments welcome.

The generated documentation now appears [here](http://inspire-schemas.readthedocs.io/en/docstests/schemas/records/journals.html).

@inspirehep/inspire-content (in particular @fschwenn and @cleggm1)

This is a continuation of PR #215.

### DIFF
--- a/inspire_schemas/records/elements/reference.yml
+++ b/inspire_schemas/records/elements/reference.yml
@@ -173,7 +173,7 @@ properties:
             parent_report_number:
                 description: |-
                     :MARC: ``999C5r`` but not distinguished from the cited
-                        document :ref:`reference.json#/properties/report_number`.
+                        document :ref:`reference.json#/properties/report_numbers`.
                 type: string
             parent_title:
                 description: |-
--- a/inspire_schemas/records/elements/related_record.yml
+++ b/inspire_schemas/records/elements/related_record.yml
@@ -9,16 +9,17 @@ properties:
     record:
         $ref: json_reference.json
         description: |-
-            :MARC: ``510__0`` (for Institutions and Experiments),
-                ``78002/78502/78708w`` (for Literature and Journals).
+            :MARC: ``510__0`` (for Institutions and Experiments), ``530__0``
+                (for Journals), ``78002/78502/78708w`` (for Literature).
         title: URL of related record
     relation:
         description: |-
             The possible values are:
 
             ``predecessor``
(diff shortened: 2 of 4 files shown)


aw-bib:
You may want to consider reusing join2 journals in your new schema. It might come in handy as you don't need to collect every value yourself if you can reuse stuff we update every year.

cf. https://bib-pubdb1.desy.de/collection/Periodicals

michamos:
Some open questions:
- What are the different types of notes? there is `640__a`, `680__i`, `667__a`, `667__x` that I am aware of. How should they be mapped to the new schema? is there any difference between them?
- `677__x` mainly contains harvesting information, that is probably outdated. Should it be discarded? there is also some historical info there, probably some confusion with `667__x`.

cleggm1:
640__a and 680__i both contain information that should be public notes, although currently only 680__i is public. Only 7 records use 667__a, so we can update them to one of the note fields that will be mapped. 667__x contains mixed information (Open Access, Online Journal, journal's description, translation of ___, info for catalogers, etc.). It seems to be used as a private note.

michamos:
@cleggm1 regarding your earlier question IRL about `related_records`, it is possible to express the relation through [`relation`](http://inspire-schemas.readthedocs.io/en/docstests/schemas/records/elements/related_record.html#relation), or [`relation_freetext`](http://inspire-schemas.readthedocs.io/en/docstests/schemas/records/elements/related_record.html#relation) in case no "standard" relation applies. Do you think more stuff is needed here?

ksachs:
https://github.com/inspirehep/inspire-schemas/blob/a3bc41140c4a15ebd8714200e93f40b74b65a7d4/inspire_schemas/records/journals.yml#L107 
there is only one journal with 980:CORE and that info is wrong.
There is no journal where all articles are considered as CORE.
I don't think we need that field.

There might be other information about journals to be used in enrichment, e.g. subject for JINST, PhysRevSTAB, see https://github.com/inspirehep/inspire-next/issues/2650

Currently hardcoded are very specific sub-sections of journals that are taken completely.
Would be nice if that could go to the journals DB too. See https://github.com/inspirehep/inspire-next/issues/2643

michamos:
:+1: I will remove core and add subjects. Not sure about how to handle the sub-sections though.

PR #220, opened 2017-08-23
Fixes the unicode author name handling.

david-caro:


jacquerie:
Whoops, good catch!

david-caro:
Not sure if it's the best way of fixing it... tested on python 2 and 3 though.

PR #222, opened 2017-08-28
Suffixes taken into account for name normalization

iulianav:
Modified the name normalization function in order to deal with suffixes in the specified way. It takes into account unicode characters as well. However, titles are still not being dealt with.

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -31,6 +31,7 @@ import six
 from jsonschema import validate as jsonschema_validate
 from jsonschema import RefResolver, draft4_format_checker
 from nameparser import HumanName
+from nameparser.config import Constants
 from pkg_resources import resource_filename
 from unidecode import unidecode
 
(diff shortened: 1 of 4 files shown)


inspire_schemas/utils.py
```diff
@@ -31,6 +31,9 @@
 from jsonschema import validate as jsonschema_validate
 from jsonschema import RefResolver, draft4_format_checker
 from nameparser import HumanName
+from nameparser.config import Constants
+from nameparser.config.capitalization import CAPITALIZATION_EXCEPTIONS
```
michamos: unused import

inspire_schemas/utils.py
```diff
@@ -31,6 +31,9 @@
...
+from nameparser.config.suffixes import SUFFIX_NOT_ACRONYMS
```
michamos: unused import

inspire_schemas/utils.py
```diff
@@ -221,21 +224,42 @@ def normalize_author_name(author):
...
+                              u'xii', u'xiii', u'xiv', u'xv']
+    for suff in roman_numeral_suffixes:
+        constants.suffix_not_acronyms.add(suff)
+
+    def _is_initial_or_dotted_suffix(author_name):
+        return len(author_name) == 1 or u'.' in author_name
 
     def _ensure_dotted_initials(author_name):
-        if _is_initial(author_name) and '.' not in author_name:
+        if _is_initial_or_dotted_suffix(author_name)\
```
michamos: why did you change the name of this function if you only call it for initials?

inspire_schemas/utils.py
```diff
@@ -221,21 +224,42 @@ def normalize_author_name(author):
...
+        if u'.' not in author_suffix:
+            seq = (author_suffix, u'.')
+            author_suffix = u''.join(seq)
+        return author_suffix
+
+    def _is_roman_numeral(numeral):
+        """Controls that the userinput only contains valid roman numerals"""
+        validRomanNumerals = [u'M', u'D', u'C', u'L', u'X',
+                              u'V', u'I', u'(', u')']
+        return all([letters in validRomanNumerals
```
michamos: you don't need the `[...]`, `all` can take an iterator.

inspire_schemas/utils.py
```diff
@@ -221,21 +224,42 @@ def normalize_author_name(author):
...
         return author_name
 
-    name = HumanName(author)
+    def _ensure_dotted_suffixes(author_suffix):
+        if u'.' not in author_suffix:
+            seq = (author_suffix, u'.')
+            author_suffix = u''.join(seq)
+        return author_suffix
+
+    def _is_roman_numeral(numeral):
```
michamos: you don't know whether the argument is a numeral as that's what you are trying to determine. You might want to call it `suffix` or something like that.

inspire_schemas/utils.py
```diff
@@ -245,9 +269,14 @@ def _ensure_dotted_initials(author_name):
...
+    if _is_roman_numeral(name.suffix):
+        suffix = name.suffix.upper()
+    else:
+        suffix = _ensure_dotted_suffixes(name.suffix)
+
     normalized_names.strip()
     final_name = u', '.join(
-        part for part in (name.last, normalized_names) if part
-    ).strip()
+        part for part in (name.last, normalized_names.rstrip(), suffix)
```
michamos: you are already `strip`ping in https://github.com/inspirehep/inspire-schemas/pull/222/files#diff-1d0970a763dbbbe297382c68c9c55ac8L248, why are you `strip`ping again?

tests/unit/test_utils.py
```diff
@@ -203,12 +203,14 @@ def test_split_pubnote():
 
 
 def test_normalize_author_name_full():
+
```
michamos: this is wrong (also in other tests). The body of the function should start on the first line.

tests/unit/test_utils.py
```diff
@@ -256,3 +263,34 @@ def test_normalize_author_name_handles_unicode():
     expected = u'蕾拉'
 
     assert expected == utils.normalize_author_name(u'蕾拉')
+
+
+def test_normalize_author_name_handles_suffixes():
```
michamos: those are 4 different tests. It's not clear to me whether you need all of them. After removing the useless ones, you may split them by hand or use [parametrize](https://docs.pytest.org/en/latest/parametrize.html) (as you wish).

PR #225, opened 2017-09-15
Change 'submission_number' field to 'holdingpen_record'

iulianav:
Because the 'submission_number' field is being used also in non-submissions,
I have renamed it in a more representative way. Soon, it will contain a
json_reference to the holdingpen record, thus the schema must be updated
accordingly.

Signed-off-by: Iuliana Voinea <iuliana.voinea@student.manchester.ac.uk>

### DIFF
--- a/tests/integration/fixtures/expected_data_hep.yaml
+++ b/tests/integration/fixtures/expected_data_hep.yaml
@@ -120,7 +120,7 @@
       "email":"albert.einstein@hep.edu",
       "source":"submitter",
       "internal_uid":1,
-      "submission_number":"12",
+      "holdingpen_record":"/holdingpen/12",
       "orcid":"0000-0001-8528-2091",
       "method":"batchuploader"
    },
--- a/tests/integration/fixtures/input_data_hep.yaml
+++ b/tests/integration/fixtures/input_data_hep.yaml
@@ -36,7 +36,7 @@
    "report_number":"7892",
    "collaboration":"Mexico City U., Cuajimalpa",
    "method":"batchuploader",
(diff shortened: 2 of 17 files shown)


tests/integration/fixtures/expected_data_hep.yaml
```diff
@@ -120,7 +120,7 @@
       "email":"albert.einstein@hep.edu",
       "source":"submitter",
       "internal_uid":1,
-      "submission_number":"12",
+      "holdingpen_record":"/holdingpen/12",
```
iulianav: @michamos I don't know what json_reference to use here

tests/integration/fixtures/input_data_hep.yaml
```diff
@@ -36,7 +36,7 @@
    "report_number":"7892",
    "collaboration":"Mexico City U., Cuajimalpa",
    "method":"batchuploader",
-   "submission_number":"12",
+   "holdingpen_record":"/holdingpen/12",
```
iulianav: @michamos also here missing json_reference

iulianav:
This should not be merged until hepcrawl and inspire-next are also modified to use the new field.

michamos:
@iulianav updating the schemas doesn't require to update first the other packages, all that is required is to bump the major version for this kind of incompatible changes, so nothing breaks as they depend on a fixed major version of `inspire-schemas`.

The comment that promped this PR was not meant to make you work on it, but mainly to write it down.
The way you did it is not correct, as you only changed the name of the field, but its type should be changed too.

PR #228, opened 2017-09-20
builders: add missing fields to make_author

michamos:
* Some of the fields that an author could have were missing from the
builder, now they are present.
* Also adds missing schema validation to some LiteratureBuilder unit
tests.

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -209,53 +209,97 @@ class LiteratureBuilder(object):
         """
         self._append_to('authors', author)
 
-    def make_author(self, full_name, affiliations=None, roles=None,
-                    raw_affiliations=None, source=None):
+    @filter_empty_parameters
+    def make_author(self, full_name,
+                    affiliations=None,
+                    roles=None,
+                    raw_affiliations=None,
+                    source=None,
+                    ids=None,
+                    emails=None,
+                    alternative_names=None):
         """Make a subrecord representing an author.
 
         Args:
             full_name(str): full name of the author. If not yet in standard
                 Inspire form, it will be normalized.
-            affiliations(list): Inspire normalized affiliations of the author.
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/literature.py
```diff
@@ -209,53 +209,97 @@ def add_author(self, author):
...
+                filter(bool, alternative_names)
+            )
+
+        def _add_inspire_roles(author, inspire_roles):
+            if inspire_roles is None:
+                return
 
-        if affiliations:
-            author = _add_affiliations(author, affiliations)
+            author.setdefault('inspire_roles', []).extend(filter(bool, roles))
```
michamos: This is wrong, if `roles=['']` it will do the wrong thing. I will change the logic.

michamos: actually, that's not true, the `filter(bool, roles)` would remove it as `''` is falsy.

tests/unit/test_literature_builder.py
```diff
@@ -178,15 +179,60 @@ def test_add_document():
...
-            'material': 'material',
+            'material': 'preprint',
             'original_url': 'original_url',
             'source': 'source',
             'url': 'url',
         },
     ]
     result = builder.record
 
+    assert validate(result['documents'], subschema) is None
```
david-caro: The assert here is nor really needed, as we don't really care what's the return value of the function, just running the validate is enough (in case of error, it will throw an exception). But I agree that it might look nicer. Up to you.

michamos: that's how it's done in `inspire-dojson`, I did the same here for consistency.

jacquerie: That's quite true though, I wasted quite a few bytes...

david-caro: That could even have some secret meaning went decoded with the right standard ;)

PR #229, opened 2017-09-20
literature: add subtitle parameter to add_title

david-caro:


david-caro:
We are populating the subtitle for the proceeding record for PoS spider, so we need this, that was missing.

david-caro:
Interesting that the fixture record is in catalan :)

PR #231, opened 2017-09-22
builders: add set_curated to LiteratureBuilder

michamos:
* Also slightly tweaks the setters for the other flags.


michamos:
I have the same doubts as in https://github.com/inspirehep/inspire-dojson/pull/94#pullrequestreview-64500001 about compatibility.

PR #233, opened 2017-09-26
Improve date handling

michamos:


inspire_schemas/utils.py
```diff
@@ -278,3 +282,90 @@ def _is_roman_numeral(suffix):
...
+
+    If parsing of textual data is needed, :func:`normalize_date` should be used
+    instead.
+
+    Returns:
+        str: a formatted date, in the form YYYY-MM-DD, YYYY-MM or YYYY
+            (depending on the information present in the date).
+
+    Raises:
+        ValueError: when year does not contain valid data.
```
michamos: Not sure about that, I think it is more pythonic to raise an exception instead of returning an empty string.

inspire_schemas/utils.py
```diff
@@ -278,3 +282,90 @@ def _is_roman_numeral(suffix):
...
+
+        >>> from inspire_schemas.utils import format_date
+        >>> format_date(*'2014-05-00'.split('-'))
+        '2014-05'
+    """
+    # XXX: 0 is not a valid year/month/day
+    non_empty = itertools.takewhile(
+        bool, (maybe_int(part) for part in (year, month, day))
+    )
+    # XXX: this only handles dates after 1000, which should be sufficient
```
michamos: it can be done, but it becomes a lot more messy for no real benefit.

inspire_schemas/builders/literature.py
```diff
@@ -330,7 +330,10 @@ def add_book(self, publisher=None, place=None, date=None):
 
         imprint = {}
         if date is not None:
-            imprint['date'] = date
+            try:
+                imprint['date'] = normalize_date(date)
+            except ValueError:
+                pass
```
michamos: maybe it should be logged instead of ignored? `inspire-schemas` does not have any logging currently, so I did it this way.

michamos:
Looking at what operations on dates are done in inspire, it looks like what is really needed is a `PartialDate` class that inherits from tuple (to get date comparison for free) and can serialize/deserialize to/from the format we use in the schemas and additionally can parse a string with `dateutil` (like in `normalize_date`) and pretty print itself to be displayed in the views. I didn't implement that as it requires a bit more work and am not completely sure about the design or whether it belongs here or in `inspire-utils`.

michamos:
superseded by inspirehep/inspire-utils#9.

PR #234, opened 2017-09-28
elements/arxiv_categories: update categories

michamos:
* INCOMPATIBLE Removes some categories that have been subsumed into
another one, which are taken care of by inspirehep/inspire-dojson#22.
* Adds the new `econ.EM` category.


kaplun:
How does it work for old records that were in categories you remove?

michamos:
as the commit message says, the category is normalized by dojson:
https://github.com/inspirehep/inspire-dojson/blob/master/inspire_dojson/utils/arxiv.py#L32-L51

PR #239, opened 2017-10-09
utils: improve classify_field

michamos:

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -79,7 +79,6 @@ _NEW_CATEGORIES = {
 }
 
 ARXIV_TO_INSPIRE_CATEGORY_MAPPING = {
-    'alg-geom': 'Math and Math Physics',
     'astro-ph': 'Astrophysics',
     'astro-ph.CO': 'Astrophysics',
     'astro-ph.EP': 'Astrophysics',
@@ -138,7 +137,6 @@ ARXIV_TO_INSPIRE_CATEGORY_MAPPING = {
     'cs.SE': 'Computing',
     'cs.SI': 'Computing',
     'cs.SY': 'Computing',
-    'dg-ga': 'Math and Math Physics',
     'gr-qc': 'Gravitation and Cosmology',
     'hep-ex': 'Experiment-HEP',
     'hep-lat': 'Lattice',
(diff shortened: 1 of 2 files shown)


inspire_schemas/utils.py
```diff
@@ -196,37 +193,7 @@
...
-    'q-fin.TR': 'Other',
     'quant-ph': 'General Physics',
-    'solv-int': 'Math and Math Physics',
-    'stat': 'Other',
-    'stat.AP': 'Other',
-    'stat.CO': 'Other',
-    'stat.ME': 'Other',
-    'stat.ML': 'Other',
-    'stat.OT': 'Other',
-    'stat.TH': 'Other',
```
jacquerie: Maybe we need a test to make sure that we're covering all the arXiv categories that we allow for in our schemas?

inspire_schemas/utils.py
```diff
@@ -268,20 +235,29 @@ def valid_arxiv_categories():
 
 
 def classify_field(value):
-    """Translate an arXiv category to the corresponding INSPIRE category.
+    """Normalize ``value`` to an Inspire category.
```
jacquerie: That's quite the nit, but I'm under the impression that INSPIRE is the correct capitalization.

inspire_schemas/utils.py
michamos: it's the other way around. The default is `Other`, because we expect the new inspire categories to be non-physics/math. This allows to add them to the schemas but not these mappings, and that's why I removed all the top-level `Other` categories. If you think that's a bad idea, we can do it the other way around.

inspire_schemas/utils.py
michamos: I don't know. I thought `INSPIRE` == `Legacy` and `Inspire` == `the future`?

inspire_schemas/utils.py
```diff
@@ -268,20 +235,29 @@ def valid_arxiv_categories():
...
 
     """
-    if isinstance(value, six.string_types):
-        for name, category in six.iteritems(ARXIV_TO_INSPIRE_CATEGORY_MAPPING):
-            if value.upper() in [name.upper(), category.upper()]:
-                return category
+    if not (isinstance(value, six.string_types) and value):
+        return
+
+    schema = load_schema('elements/inspire_field')
```
jacquerie: This reminds me: we need to fix `load_schema` to resolve schema references, so that we can express schema paths relative to the root and not to the arbitrary separations in elements (but of course, not in this PR.)

PR #242, opened 2017-10-12
Builder improvements

michamos:
Closes inspirehep/hepcrawl#176.

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -194,7 +194,8 @@ class LiteratureBuilder(object):
         :param material: material for the doi.
         :type material: string
         """
-        if not idutils.normalize_doi(doi):
+        doi = idutils.normalize_doi(doi)
+        if not doi:
             return
 
         dois = self._sourced_dict(
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -31,7 +31,7 @@ from functools import wraps
 
 import idutils
 
-from ..utils import normalize_author_name, validate
+from ..utils import normalize_author_name, normalize_collaboration, validate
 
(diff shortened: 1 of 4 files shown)


inspire_schemas/builders/literature.py
```diff
@@ -48,7 +48,8 @@ def func_wrapper(self, *args, **kwargs):
                      if value not in EMPTIES}
 
         if (
-            list(my_kwargs.keys()) == ['source'] or not list(my_kwargs.keys())
+            set(my_kwargs.keys()).issubset(['source', 'material']) or
+                not list(my_kwargs.keys())
```
kaplun: :boom:

kaplun: You can slightly simplify with `(set(['source', 'material']) - set(my_kwargs)) or not my_kwargs`.

inspire_schemas/builders/literature.py
```diff
@@ -194,7 +195,8 @@ def add_doi(self, doi, source=None, material=None):
         :param material: material for the doi.
         :type material: string
         """
-        if not idutils.normalize_doi(doi):
+        doi = idutils.normalize_doi(doi)
+        if not doi:
```
kaplun: This seems to fix a bug that is not documented in the commit log though.

michamos: :+1: forgot about that one. Will also add a test.

inspire_schemas/builders/literature.py
michamos: actually that doesn't work. I want it to be true when `my_kwargs` is a subset of `{'source', 'material'}`, which is not the case for your proposal (or its negation).

kaplun: That's why the :boom: :smile:

PR #243, opened 2017-10-12
reference_builder: use `normalize_author_name`.

ammirate:
Signed-off-by: Antonio Cesarano <cesarano2607@gmail.com>

Closes inspirehep/inspire-dojson#69

### DIFF
--- a/tests/unit/test_reference_builder.py
+++ b/tests/unit/test_reference_builder.py
@@ -31,50 +31,6 @@ from inspire_schemas.builders.references import (
 )
 
 
-def test_split_refextract_authors_str_initials():
-    expected = [
-        'Butler, D.',
-        'Demarque, P.',
-        'Smith, H.A.',
-    ]
-    authors_input = 'D. Butler, P. Demarque, & H. A. Smith'
-    result = _split_refextract_authors_str(authors_input)
-
-    assert expected == result
-
-
-def test_split_refextract_authors_str_noninitials():
-    expected = [
-        'Klebanov, Igor R.',
-        'Maldacena, Juan Martin'
-    ]
-    authors_input = 'Igor R. Klebanov and Juan Martin Maldacena'
-    result = _split_refextract_authors_str(authors_input)
-
-    assert expected == result
-
-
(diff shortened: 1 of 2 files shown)


michamos:
actually only `add_author` seems better, so that names get normalized no matter the origin.

ammirate:
You mean not using `normalize_author_name` then extracting names in `_split_refextract_authors_str`, but only in `add_author`?

ammirate:
@michamos do the tests look meaningful for you?

tests/unit/test_reference_builder.py
```diff
@@ -452,6 +408,82 @@ def test_add_refextract_author_str():
...
+
+    builder.add_refextract_authors_str(
+        'S. B. Cenko, M. M. Kasliwal, D. A. Perley et al.'
+    )
+
+    expected = [
+        {
+            'reference': {
+                'authors': [
+                    {'full_name': 'Cenko, S. B.'},
```
michamos: this is strange, it should remove the spaces:
```
In [1]: from inspire_schemas.utils import normalize_author_name

In [2]: normalize_author_name('Cenko, S. B.')
Out[2]: u'Cenko, S.B.'
```

tests/unit/test_reference_builder.py
```diff
@@ -452,6 +408,82 @@ def test_add_refextract_author_str():
...
+            },
+        },
+    ]
+    result = [builder.obj]
+
+    assert validate(result, subschema) is None
+    assert expected == result
+
+
+def test_add_refextract_authors_str_discards_et_al():
```
michamos: repeated function name.

tests/unit/test_reference_builder.py
ammirate: True. Then, why the build passed? :/

michamos: becaues of the other issue, you repeated the function name, so you redefined the function, so this test never ran.

ammirate: fair enough

PR #244, opened 2017-10-17
builders: get license abbreviation from URL

michamos:


inspire_schemas/utils.py
```diff
@@ -551,3 +555,34 @@ def normalize_collaboration(collaboration):
...
+    Raises:
+        ValueError: when the url is not recognized
+    """
+    if not url:
+        return
+
+    split_url = urlsplit(url, scheme='http')
+
+    if split_url.netloc.lower() == 'creativecommons.org':
+        license = ['CC']
```
jacquerie: Fun fact: `license` is actually shadowing a function in the Python interpreter (https://docs.python.org/2/library/constants.html#constants-added-by-the-site-module). But we don't really care, it's useless...

PR #249, opened 2017-10-25
utils: relax artid decision in split_page_artid

michamos:


jacquerie:
That function really needs a refactoring, though, and probably a lot more tests.

PR #252, opened 2017-10-27
hep.documents: make url required

david-caro:
The url is a field that must always be populated, either with the url
to download the document from, or with the api path to the downloaded
resource.


jacquerie:
Note however that is an INCOMPATIBLE change, as records that satisfied an older schema might not be valid in this new one.

michamos:
@jacquerie taking care of that

david-caro:
Superseded by #253

PR #253, opened 2017-10-30
Require url and validate date-time format

michamos:


michamos:
This is one PR so that we don't generate two major versions for these two "small" incompatible changes.

PR #255, opened 2017-11-02
global: make values required

michamos:
* INCOMPATIBLE Make some values required when it doesn't make sense to
have the field without them.
* Also removes ``undefined`` from ``inspire_field.source``.


michamos:
superseded by #105.

PR #256, opened 2017-11-02
utils: handle journal letter split

jacquerie:
https://www.youtube.com/watch?v=VKcAYMb5uk4

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE-SCHEMAS.
-# Copyright (C) 2016 CERN.
+# Copyright (C) 2016, 2017 CERN.
 #
 # INSPIRE-SCHEMAS is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
@@ -23,6 +23,8 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 """Public api for methods and functions to handle/verify the jsonschemas."""
+
+import copy
 import json
 import os
 import re
@@ -53,6 +55,15 @@ _RE_COLLABORATION_TRAILING = re.compile(
 )
 _RE_LICENSE_URL = re.compile(
     r'^/licenses/(?P<sublicense>[-\w]*)(?:/(?P<version>[\.\d]*))?'
+)
+_RE_VOLUME_STARTS_WITH_A_LETTER = re.compile(
+    r'^(?P<letter>[A-Z])(?P<volume>\d+)', re.IGNORECASE
+)
+_RE_VOLUME_ENDS_WITH_A_LETTER = re.compile(
+    r'(?P<volume>\d+)(?P<letter>[A-Z])$', re.IGNORECASE
+)
+_RE_TITLE_ENDS_WITH_A_LETTER = re.compile(
+    r'(?P<title>.+(\.| ))(?P<letter>[A-Z])$', re.IGNORECASE
 )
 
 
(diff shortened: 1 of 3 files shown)


inspire_schemas/utils.py
```diff
@@ -209,6 +218,54 @@
     'quant-ph': 'General Physics',
 }
 
+_SPECIAL_CASES = [
```
michamos: name too generic, maybe `_JOURNALS_WITH_LETTERS`?

inspire_schemas/utils.py
```diff
@@ -209,6 +218,54 @@
     'quant-ph': 'General Physics',
 }
 
+_JOURNALS_WITH_ALREADY_A_LETTER = [
```
michamos: shouldn't this be a `set`?

inspire_schemas/utils.py
jacquerie: I had already changed it to `_JOURNALS_WITH_ALREADY_A_LETTER` without reading your comment : )

inspire_schemas/utils.py
jacquerie: Sure, although IIRC there's not a lot of performance difference with containers of this size.

inspire_schemas/utils.py
```diff
@@ -209,6 +220,58 @@
...
+    'Proc.Rom.Acad.A',
+    'Rev.Univ.Nac.Tucuman, Ser.A',
+    'Sci.Rep.Nat Tsing Hua Univ.Ser.A',
+    'Spectrochim.Acta A',
+    'Tellus A',
+    'Trans.Int.Astron.Union A',
+])
+
+_JOURNALS_THAT_NEED_A_HIDDEN_PUBNOTE = {
+    'Phys.Lett.B': set([str(el) for el in range(24, 171) if el != 54]),
```
michamos: I wouldn't do the `54` exception: the reason it's not needed is because we don't have records, not because it's in principle not needed. You can also drop the `[...]` as `set` can take an iterable.

inspire_schemas/utils.py
```diff
@@ -568,3 +631,61 @@ def get_license_from_url(url):
...
+    for publication_info in publication_info_list:
+        _publication_info = copy.deepcopy(publication_info)
+
+        hidden = publication_info.get('hidden')
+        journal_title = publication_info.get('journal_title')
+        journal_volume = publication_info.get('journal_volume')
+        if journal_title and journal_volume:
+            match = _RE_VOLUME_STARTS_WITH_A_LETTER.match(journal_volume)
+            if match:
+                if 'journal_record' in _publication_info:
```
michamos: `_publication_info.pop('journal_record', None)` is a one-liner. Maybe a bit less explicit though...

inspire_schemas/utils.py
```diff
@@ -568,3 +631,61 @@ def get_license_from_url(url):
...
+                    del _publication_info['journal_record']
+                if journal_title.endswith('.'):
+                    _publication_info['journal_title'] = journal_title + match.group('letter')
+                else:
+                    _publication_info['journal_title'] = journal_title + ' ' + match.group('letter')
```
michamos: `u''.join`/`u' '.join` here instead?

inspire_schemas/utils.py
```diff
@@ -568,3 +631,61 @@ def get_license_from_url(url):
...
+
+def convert_old_publication_info_to_new(publication_info_list):
+    result = []
+
+    for publication_info in publication_info_list:
+        _publication_info = copy.deepcopy(publication_info)
+
+        hidden = publication_info.get('hidden')
+        journal_title = publication_info.get('journal_title')
+        journal_volume = publication_info.get('journal_volume')
```
michamos: maybe use `_publication_info` everywhere just to be extra-safe, like you are doing in new to old?

jacquerie:
> :shipit:

:ship:

PR #261, opened 2017-11-08
setup:resolve schemas on build time

Glignos:
Signed-off-by: Glignos <glignos93@gmail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,19 +30,47 @@ import os
 
 from setuptools import find_packages, setup
 
+import yaml
+
 URL = 'https://github.com/inspirehep/inspire-schemas'
 
 
+def _resolve_json_schema(json_schema, path):
+    if isinstance(json_schema, list):
+        json_schema = [_resolve_json_schema(item, path) for item in json_schema]
+    elif isinstance(json_schema, dict):
+        for key in json_schema:
+            if key == '$ref' and not isinstance(json_schema[key], dict):
+                subschema_path = os.path.join(path, json_schema[key])
+                subschema_path = os.path.splitext(subschema_path)[0]
+                with open(subschema_path+'.yml', 'rb') as yaml_fd:
+                    raw_data = yaml_fd.read()
+                data = yaml.load(raw_data)
+                return data
+            else:
+                json_schema[key] = _resolve_json_schema(json_schema[key], path)
+
+    return json_schema
+
+
 def _yaml2json(yaml_file, json_file):
-    import yaml
     with open(yaml_file, 'rb') as yaml_fd:
         raw_data = yaml_fd.read()
-
     data = yaml.load(raw_data)
-
     with open(json_file, 'w') as json_fd:
-        json_fd.write(
-            json.dumps(data, indent=4, separators=(',', ': '), sort_keys=True)
+        json.dump(
+            data, json_fd, indent=4, separators=(',', ': '), sort_keys=True
+        )
+        json_fd.write('\n')
+
+    path = os.path.dirname(yaml_file)
+    resolved_json_object = _resolve_json_schema(data, path)
+    resolved_record = json_file.replace('records', 'resolved_records')
+    if not os.path.isdir(os.path.dirname(resolved_record)):
+        os.mkdir(os.path.dirname(resolved_record))
+    with open(resolved_record, 'w') as json_fd:
+        json.dump(
+            resolved_json_object, json_fd, indent=4, separators=(',', ': '), sort_keys=True
         )
         json_fd.write('\n')
 
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 # JSON files, generated from YAML
 inspire_schemas/records/**/*.json
+inspire_schemas/resolved_records/**/*.json
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
@@ -78,6 +79,7 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
+docs/schemas/
 
 # PyBuilder
 target/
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -30,20 +30,39 @@
...
+
 def _yaml2json(yaml_file, json_file):
-    import yaml
     with open(yaml_file, 'rb') as yaml_fd:
         raw_data = yaml_fd.read()
 
     data = yaml.load(raw_data)
-
+    json_object = json.loads(json.dumps(data, indent=4, separators=(',', ': '), sort_keys=True))
+    path = os.path.dirname(yaml_file) + '/'
```
michamos: you don't have to add the `/` by hand if you use `os.path.join` as suggested earlier.

setup.py
```diff
@@ -30,20 +30,39 @@
...
+
+
 def _yaml2json(yaml_file, json_file):
-    import yaml
     with open(yaml_file, 'rb') as yaml_fd:
         raw_data = yaml_fd.read()
 
     data = yaml.load(raw_data)
-
+    json_object = json.loads(json.dumps(data, indent=4, separators=(',', ': '), sort_keys=True))
```
michamos: again, dumping and the loading does nothing.

setup.py
```diff
@@ -30,20 +30,39 @@
...
+    path = os.path.dirname(yaml_file) + '/'
+    resolved_json_object = _resolve_json_schema(json_object, path)
     with open(json_file, 'w') as json_fd:
-        json_fd.write(
-            json.dumps(data, indent=4, separators=(',', ': '), sort_keys=True)
-        )
+        json_data = json.dumps(resolved_json_object, indent=4, separators=(',', ': '), sort_keys=True)
+        json_fd.write(json_data)
```
michamos: you can do `json.dump(resolved_json_object, json_fd, ...)` instead of having these two separate lines.

setup.py
```diff
@@ -30,20 +30,39 @@
...
 from setuptools import find_packages, setup
 
+import yaml
+
 URL = 'https://github.com/inspirehep/inspire-schemas'
 
 
+def _resolve_json_schema(json_schema, path):
+    if isinstance(json_schema, list):
+        json_schema[0] = _resolve_json_schema(json_schema[0], path)
```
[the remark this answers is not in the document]
Glignos: I think that when there is an array it's length is always 1 since it contains only the properties of an object or a value.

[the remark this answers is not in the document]
michamos: :confused:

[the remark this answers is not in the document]
Glignos: I guess in the case of AnyOf that it's needed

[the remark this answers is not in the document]
michamos: I guess so too, but then why only resolving the first element of `anyOf`?

setup.py
Glignos: If I try to json.loads() the yaml without dumping it first I get a decode error

setup.py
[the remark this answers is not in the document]
Glignos: Yes I meant that in that case we need indeed a loop :)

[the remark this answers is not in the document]
michamos: or rather a list comprehension

Glignos:
@michamos check again? After a talk with @jacquerie I created a separate directory for the resolved ones.

setup.py
```diff
@@ -30,21 +30,55 @@
...
-
     with open(json_file, 'w') as json_fd:
-        json_fd.write(
-            json.dumps(data, indent=4, separators=(',', ': '), sort_keys=True)
-        )
+        json_fd.write(json.dumps(
+            data, indent=4, separators=(',', ': '), sort_keys=True
+        ))
+        json_fd.write('\n')
+        json_fd.close()
```
michamos: no need to `close` manually, the context manager you are in is handling that for you.

setup.py
```diff
@@ -30,21 +30,55 @@
...
+
+    path = os.path.dirname(yaml_file)
+    resolved_json_object = _resolve_json_schema(data, path)
+    resolved_record = json_file.replace('records', 'resolved_records')
+    try:
+        os.stat(os.path.dirname(resolved_record))
```
michamos: this is a bit convoluted. Why not trying to `os.mkdir` and `except OSErrror: pass` instead?

setup.py
```diff
@@ -30,21 +30,55 @@
...
     data = yaml.load(raw_data)
-
     with open(json_file, 'w') as json_fd:
-        json_fd.write(
-            json.dumps(data, indent=4, separators=(',', ': '), sort_keys=True)
-        )
+        json_fd.write(json.dumps(
+            data, indent=4, separators=(',', ': '), sort_keys=True
+        ))
+        json_fd.write('\n')
```
michamos: you can replace these two lines with `json.dump(data, json_fd, ...)`

setup.py
```diff
@@ -30,21 +30,55 @@
...
+    try:
+        os.stat(os.path.dirname(resolved_record))
+    except:
+        os.mkdir(os.path.dirname(resolved_record))
+    with open(resolved_record, 'w') as json_fd:
+        json_fd.write(json.dumps(
+            resolved_json_object, indent=4, separators=(',', ': '), sort_keys=True
+        ))
         json_fd.write('\n')
+        json_fd.close()
```
michamos: same comments as for the other case.

setup.py
```diff
@@ -98,6 +131,7 @@ def do_setup():
         zip_safe=False,
         autosemver=True,
     )
+    _generate_json_schemas()
```
michamos: this was correct previously, now the package is installable but it will not generate the schemas when creating the egg.

Glignos:
@michamos Did the changes check again? :)

Glignos:
@michamos did the squash :)

michamos:
@Glignos I force-pushed on your branch to fix some small styling issues, so be careful.

PR #263, opened 2017-11-10
builders: add journal related setters for references

szymonlopaciuk:
To `ReferenceBuilder` add functionality to set:

- journal title, issue and volume
- start and end page and/or article id

Currently it's possible to do a workaround using `build_pubnote` and `set_pubnote`, however it's not possible to do that for `journal_issue` field. This adds above facilities to add the fields directly.

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -146,6 +146,11 @@ class ReferenceBuilder(object):
         if field_name not in self.obj['reference']:
             self.obj['reference'][field_name] = value
 
+    def _set_publication_info_field(self, field_name, value):
+        """Put a value in the publication info of the reference."""
+        self._ensure_reference_field('publication_info', {})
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/references.py
```diff
@@ -278,3 +283,32 @@ def add_uid(self, uid):
...
+
+    def set_journal_issue(self, journal_issue):
+        """Add journal issue."""
+        self._set_publication_info_field('journal_issue', journal_issue)
+
+    def set_journal_volume(self, journal_volume):
+        """Add journal volume."""
+        self._set_publication_info_field('journal_volume', journal_volume)
+
+    def set_page_artid(self, page_start, page_end, artid):
```
michamos: can you change this to `set_page_artid(self, page_start=None, page_end=None, artid=None):` and raise a `ValueError` when there is a `page_end` but no `page_start`? this allows to provide only a (valid) subset of data.

PR #264, opened 2017-11-13
fix UnicodeEncodeError when writing to the changelog

szymonlopaciuk:
Replaced open with codecs.open to write unicode to the `CHANGELOG.txt` file as UTF-8.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,6 +25,7 @@
 from __future__ import print_function
 
 import os
+from io import open
 
 from autosemver.packaging import get_changelog, get_current_version
 from jsonschema2rst.parser_runner import run_parser


docs/conf.py
```diff
@@ -25,6 +25,7 @@
 from __future__ import print_function
 
 import os
+import codecs
```
jacquerie: You should be able to do `from io import open` here and then you won't have to touch the code below. `codecs` only matters when you have to support Python 2.5 (which is never).

szymonlopaciuk: Done in d7cd775

michamos:
can you rebase on master? I just merged #261.

docs/conf.py
jacquerie: > Done in d7cd775

Something went wrong: this commit still contains the old code.

jacquerie: Ok, you pushed be2a5fc to fix this.

szymonlopaciuk: yup, messed up. Should be fine now

szymonlopaciuk:
@michamos rebased 👍

PR #265, opened 2017-11-13
elements/id: require ORCIDs to be valid

michamos:
* As suggested by @tsgit, this adds a new `orcid` `format` that uses
`idutils.is_orcid` for validation (as well as the corresponding example
generator) to ensure that valid records can contain only valid ORCIDs.

### DIFF
--- a/inspire_schemas/records/elements/id.yml
+++ b/inspire_schemas/records/elements/id.yml
@@ -67,6 +67,7 @@ anyOf:
         value:
             description: |-
                 :example: ``0000-0012-1234-5647``
+            format: orcid
             pattern: ^\d{4}-\d{4}-\d{4}-\d{3}[0-9X]$
             type: string
     required:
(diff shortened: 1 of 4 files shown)


michamos:
@jacquerie this is formally incompatible, as we are now more strict about ORCIDs, but shouldn't have any effect (except for potentially failing migration of some records), so I didn't mark it as incompatible in order to avoid the version-bump dance.

tsgit:
I did a major cleanup on legacy, there are about 5 Hep records and 3 HepNames records with invalid ORCID left. Heath is going to fix those. So migration from legacy shouldn't cause problems with this

jacquerie:
> this is formally incompatible, as we are now more strict about ORCIDs, but shouldn't have any effect

It might break the build of `inspire-next`, as [this](https://github.com/inspirehep/inspire-next/blob/489a861d7030e778c03daa256f6b52654dcf901d/tests/acceptance/authors/test_authors_new_form.py#L39) is the kind of ORCIDs we use in tests...

PR #268, opened 2017-11-14
builders: detect APS new-style arXiv ID with class

szymonlopaciuk:
Detect and normalise arXiv IDs of the form `arXiv:hep-th/1601.07616`, which are technically invalid, but appear in APS records.


inspire_schemas/builders/references.py
```diff
@@ -40,6 +40,12 @@
 # Matches any stream of initials (A. B C D. -E F).
 RE_INITIALS_ONLY = re.compile(r'^\s*-?[A-Z]((\.|\s)\s*-?[A-Z])*\.?\s*$',
                               re.U)
+# Matches new style arXiv ID, with an old-style class specification
+# (Malformed, but appears in APS records)
+RE_ARXIV_POST_2007_CLASS = re.compile(
+    "(arxiv:)(?:[a-z\-]+)(?:\.[a-z]{2})?/(\d{4})\.(\d{4,5})(v\d+)?$",
```
michamos: `(arxiv:)` should be optional, I would not be surprised to see also `hep-th/1501.0001`.

PR #275, opened 2017-11-20
global: use uri/uri-reference formats, not url

### LINKED ISSUE #274: documents -> url - remove url format?
The field documents->url is defined as url but currently we also store paths (which makes validation on the editor fail)

screen shot 2017-11-20 at 9 59 53 am

michamos:
* INCOMPATIBLE This replaces the `url` `format` everywhere by `uri` (for
absolute URLs) or `uri-reference` (if the URL could be relative).  These
`format`s are part of the JSON schema spec (the second one starting from
Draft 6), whereas `url` is not. In order to validate these formats, this
package now depends on `rfc3987`, and adds a custom validator for
`uri-reference`, as we are currently on Draft 4 (closes #274).

Note that the `json-schema-faker` currently generates relative URIs for
`uri` (which is incompatible with Draft 6), so the fixtures had to be
amended by hand.

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -28,7 +28,9 @@ import copy
 import json
 import os
 import re
+from functools import partial
 
+import rfc3987
 import six
 from inspire_utils.date import PartialDate
 from jsonschema import validate as jsonschema_validate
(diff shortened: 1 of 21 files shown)


inspire_schemas/utils.py
```diff
@@ -509,6 +511,9 @@ def load_schema(schema_name, resolved=False):
 
 inspire_format_checker = draft4_format_checker
 inspire_format_checker.checks('date', raises=ValueError)(PartialDate.loads)
+inspire_format_checker.checks('uri-reference', raises=ValueError)(
+    partial(rfc3987.parse, rule='URI_reference')
+)
```
jacquerie: (I should have ignored `E501` [here](https://github.com/inspirehep/inspire-schemas/blob/6121e80e1705f922ffab4b00a9965d3b7fd48459/setup.cfg#L64), is it not working?)

PR #276, opened 2017-11-21
builders: create a signature builder

szymonlopaciuk:
Create a builder for hep.authors records. This is a continuation of [@kaplun's work](https://github.com/kaplun/inspire-schemas/blob/signatures/inspire_schemas/builders/signatures.py)

## Motivation & Context

This is a step towards inspirehep/inspire-next#2971

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -27,36 +27,15 @@
 from __future__ import absolute_import, division, print_function
 
 import warnings
-from functools import wraps
 
 import idutils
-from inspire_utils.name import normalize_name
 
 from ..utils import (get_license_from_url,
                      normalize_collaboration,
-                     validate)
-
-EMPTIES = [None, '', [], {}]
-
-
-def filter_empty_parameters(func):
-    """Decorator that is filtering empty parameters.
-
-    :param func: function that you want wrapping
-    :type func: function
-    """
-    @wraps(func)
-    def func_wrapper(self, *args, **kwargs):
-        my_kwargs = {key: value for key, value in kwargs.items()
-                     if value not in EMPTIES}
-
-        if (
-                {'source', 'material'}.issuperset(my_kwargs) or not my_kwargs
-        ) and args == ():
-            return
-        return func(self, *args, **my_kwargs)
-
-    return func_wrapper
+                     validate,
+                     filter_empty_parameters,
+                     EMPTIES)
+from .signatures import SignatureBuilder
 
 
 def is_citeable(publication_info):
@@ -224,13 +203,13 @@ class LiteratureBuilder(object):
 
     @filter_empty_parameters
     def make_author(self, full_name,
-                    affiliations=None,
-                    roles=None,
-                    raw_affiliations=None,
+                    affiliations=(),
+                    roles=(),
+                    raw_affiliations=(),
                     source=None,
-                    ids=None,
-                    emails=None,
-                    alternative_names=None):
+                    ids=(),
+                    emails=(),
+                    alternative_names=()):
         """Make a subrecord representing an author.
 
         Args:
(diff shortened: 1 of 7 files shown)


inspire_schemas/builders/signatures.py
```diff
@@ -0,0 +1,141 @@
...
+        self.obj['ids'] = filter(lambda uid: uid.get('schema') != schema, self.obj['ids'])
+        self._add_uid(uid, schema)
+
+    @filter_empty_parameters
+    def add_inspire_role(self, inspire_role):
+        self._ensure_list_field('inspire_roles', inspire_role)
+
+    @filter_empty_parameters
+    def add_raw_affiliation(self, raw_affiliation):
+        self._ensure_list_field('raw_affiliations', raw_affiliation)
```
michamos: this looks wrong, as `raw_affiliations` elements are `sourced_value`s. You should do something similar to https://github.com/inspirehep/inspire-schemas/blob/ac809becb858cfe2a1c695c18f00014109138dca/inspire_schemas/builders/references.py#L187-L196

inspire_schemas/utils.py
```diff
@@ -293,6 +295,62 @@
...
+    Args:
+        uid (string): a UID string
+
+    Returns:
+        Tuple[string, string]: a tuple (schema, uid) where:
+        - schema: a schema of the UID or None if not recognised
+        - uid: the UID normalized to comply with the id.json schema
+    """
+    if re.compile(r'^\d{8}$').match(uid):
+        # Could be either VIAF or INSPIRE ID, give up
```
michamos: How should one add an INSPIRE ID then?

inspire_schemas/builders/signatures.py
```diff
@@ -0,0 +1,141 @@
...
+    def set_uid(self, uid):
+        """Set a unique ID.
+
+        If a UID of a given schema already exists in a record it will
+        be overwritten, otherwise it will be appended to the record.
+
+        Args:
+            uid (string): unique identifier
+        """
+        schema, uid = guess_author_uid_schema_and_normalize(uid)
```
michamos: I wouldn't do the guessing if you cannot always guess reliably. I think it's better if `set_uid` takes also `schema` as argument.

inspire_schemas/builders/signatures.py
```diff
@@ -0,0 +1,141 @@
...
+        if not schema:
+            return
+        self._ensure_field('ids', [])
+        self.obj['ids'] = filter(lambda uid: uid.get('schema') != schema, self.obj['ids'])
```
michamos: More than one identifier with the same schema is strange, but is in principle allowed, so you shouldn't filter here.

inspire_schemas/builders/signatures.py
```diff
@@ -0,0 +1,141 @@
...
+import idutils
+
+from ..utils import (
+    filter_empty_parameters,
+    guess_author_uid_schema_and_normalize,
+)
+
+
+class SignatureBuilder(object):
+    """Class used for building JSON signatures objects given simple properties.
```
michamos: maybe also explain that signatures are authors on literature records.

inspire_schemas/builders/signatures.py
szymonlopaciuk: Should I come back to what was there before, so trying to guess those things that can be guessed reliably (e.g. ORCID, DESY, maybe BAI) with an optional schema argument, or scrape the whole thing and force the user to explicitly state it?

inspire_schemas/utils.py
szymonlopaciuk: INSPIRE-12345678, but I see your point, maybe it's better to remove ambiguity

inspire_schemas/builders/signatures.py
michamos: I see your point. Let's keep it like this then, and we'll add the other case if the need ever arises.

inspire_schemas/builders/signatures.py
michamos: :+1: that would be better. Also most IDs never appear on papers in practice. I would say that you should support those that dojson handles https://github.com/inspirehep/inspire-dojson/blob/e5096bb6c78c7e52f65319fcfbe8e2325e9bb959/inspire_dojson/hep/rules/bd1xx.py#L46

inspire_schemas/builders/literature.py
```diff
@@ -224,13 +203,13 @@ def add_author(self, author):
 
     @filter_empty_parameters
     def make_author(self, full_name,
-                    affiliations=None,
-                    roles=None,
-                    raw_affiliations=None,
+                    affiliations=[],
```
michamos: you should never use mutable types in default arguments (see http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) unless you really know what you are doing. In this case, you want something iterable and immutable, so you can use the empty tuple `()` instead.

inspire_schemas/builders/signatures.py
michamos: here the guessed schema is overriding the one passed to the function. It should be the other way around.

inspire_schemas/utils.py
```diff
@@ -293,6 +295,62 @@
...
+        if (
+                {'source', 'material'}.issuperset(my_kwargs) or not my_kwargs
+        ) and args == ():
+            return
+        return func(self, *args, **my_kwargs)
+
+    return func_wrapper
+
+
+def guess_author_uid_schema_and_normalize(uid):
```
michamos: could you change this according to https://github.com/inspirehep/inspire-schemas/pull/276/files#r152972797 ?

inspire_schemas/utils.py
```diff
@@ -318,38 +318,57 @@ def func_wrapper(self, *args, **kwargs):
...
+    Raise:
+        ValueError: if specified schema is not matching the given UID
     """
-    if re.compile(r'^\d{8}$').match(uid):
-        # Could be either VIAF or INSPIRE ID, give up
-        return None, uid
+    if idutils.is_orcid(uid) and schema in (None, 'ORCID'):
+        return idutils.normalize_orcid(uid), 'ORCID'
+
+    RE_MATCH = {
```
michamos: better to have this as a global, outside of the function, so you don't rebuild this on every function call.

inspire_schemas/utils.py
```diff
@@ -293,6 +304,81 @@
...
+        raise UnknownUIDSchema(uid)
+
+    if schema:
+        normalized_uid = _get_uid_normalized_in_schema(uid, schema)
+        if normalized_uid:
+            return normalized_uid, schema
+        else:
+            raise SchemaUIDConflict(schema, uid)
+
+    match_schema, normalized_uid = None, None
```
michamos: you only need the first one for testing for duplicates (`matched_schema` sounds better), so no need to initialize both here.

inspire_schemas/utils.py
```diff
@@ -293,6 +304,81 @@
...
+    for candidate_schema in _RE_AUTHORS_UID:
+        candidate_uid = _get_uid_normalized_in_schema(uid, candidate_schema)
```
michamos: `candidate_uid` -> `normalized_uid`

inspire_schemas/utils.py
```diff
@@ -293,6 +304,81 @@
...
+        if candidate_uid:
+            if match_schema:
```
michamos: `if ... and ...:`

inspire_schemas/utils.py
```diff
@@ -293,6 +304,81 @@
...
+                # Valid against more than one candidate schema, ambiguous
+                raise UnknownUIDSchema(uid)
+            match_schema = candidate_schema
+            normalized_uid = candidate_uid
```
michamos: line not needed with previous change.

tests/unit/test_utils.py
```diff
@@ -678,3 +678,44 @@ def test_convert_new_publication_info_to_old_handles_year_added_to_volumes():
...
+])
+def test_author_id_normalize_and_schema(
+        uid, explicit_schema, expected_uid, expected_schema):
+    normalized_uid, guessed_schema = utils.author_id_normalize_and_schema(uid, explicit_schema)
+    assert guessed_schema == expected_schema
+    assert normalized_uid == expected_uid
+
+
+def test_author_id_normalize_and_schema_unknown():
+    try:
```
michamos: to assert that an exception has been raised, you should use the [`pytest.raises`](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions) context manager.

tests/unit/test_signature_builder.py
```diff
@@ -0,0 +1,226 @@
...
+
+    builder = SignatureBuilder({'existing': ['old']})
+    builder._ensure_list_field('created', 'new')
+    builder._ensure_list_field('existing', 'new')
+    result = builder.obj
+
+    assert expected == result
+
+
+def test_signature_builder(subschema):
```
michamos: you should also test the case where you initialize the builder with an existing author.

inspire_schemas/utils.py
```diff
@@ -363,18 +363,18 @@ def _get_uid_normalized_in_schema(_uid, _schema):
         else:
             raise SchemaUIDConflict(schema, uid)
 
-    match_schema, normalized_uid = None, None
+    matched_schema = None
```
michamos: I now see why you needed both now, it was better before, sorry.

Issue #279, opened 2017-11-25
elements: the $schema keyword MUST NOT appear in subschemas


jacquerie:
> The "$schema" keyword SHOULD be used in a root schema. It MUST NOT appear in subschemas.
(http://json-schema.org/latest/json-schema-core.html#rfc.section.7)

PR #281, opened 2017-11-27
elements: remove $schema from all elements

jacquerie:
Closes https://github.com/inspirehep/inspire-schemas/issues/279.

According to the spec, ``$schema`` MUST NOT appear in subschemas.
See: http://json-schema.org/latest/json-schema-core.html#rfc.section.7


michamos:
This seems to confuse `load_schema` on elements.

PR #284, opened 2017-11-27
LiteratureBuilder: fail on duplicated docs and figures

david-caro:

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -69,6 +69,14 @@ def is_citeable(publication_info):
     return has_pub_info and has_page_or_artid
 
 
+def key_already_there(element, elements):
+    """Checks if the given element existng by using the key `key`."""
+    for existing_element in elements:
+        if element['key'] == existing_element['key']:
+            return True
+    return False
+
+
 class LiteratureBuilder(object):
     """Literature record builder."""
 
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/literature.py
```diff
@@ -909,8 +917,13 @@ def add_figure(
             if locals()[dict_key] is not None:
                 figure[dict_key] = locals()[dict_key]
 
-        if figure:
-            self._append_to('figures', figure)
+        if key_already_there(figure, self.record.get('figures', ())):
+            raise ValueError(
+                'There\'s already a figure with the key %s.'
```
michamos: Is the `key` never unicode?

inspire_schemas/builders/literature.py
```diff
@@ -90,6 +90,14 @@ def _item_has_page_or_artid(item):
     return has_pub_info and has_page_or_artid
 
 
+def key_already_there(element, elements):
+    """Checks if the given element existng by using the key `key`."""
+    for existing_element in elements:
```
michamos: This is making insertion quadratic in the number of elements. Maybe better to maintain two `set`s of keys on the side.

inspire_schemas/builders/literature.py
david-caro: It does not matter:
```python
In [1]: lolo = u'sé'

In [2]: 'bom %s' % lolo
Out[2]: u'bom s\xe9'
```

inspire_schemas/builders/literature.py
david-caro: I don't really think we should care much about that now, we will have ~100 figures/documents, that's peanuts, imo not really worth the complexity of adding the sets.

PR #287, opened 2017-11-30
utils: filter_empty_parameters if all are empty

szymonlopaciuk:
This fixes an issue with `filter_empty_parameters` where if all arguments provided to a decorated function as args (as opposed to kwargs) are empty, the filtered function would still be executed on all-empty arguments.

Can be reproduced by calling `LiteratureBuilder().add_abstract(None)`.

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -318,11 +318,11 @@ def filter_empty_parameters(func):
     def func_wrapper(self, *args, **kwargs):
         my_kwargs = {key: value for key, value in kwargs.items()
                      if value not in EMPTIES}
-        my_args = [arg for arg in args if arg not in EMPTIES]
+        args_is_empty = all(arg in EMPTIES for arg in args)
 
         if (
                 {'source', 'material'}.issuperset(my_kwargs) or not my_kwargs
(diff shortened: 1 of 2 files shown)


michamos:
Looks good, could you add a test?

inspire_schemas/utils.py
```diff
@@ -318,10 +318,11 @@ def filter_empty_parameters(func):
     def func_wrapper(self, *args, **kwargs):
         my_kwargs = {key: value for key, value in kwargs.items()
                      if value not in EMPTIES}
+        my_args = [arg for arg in args if arg not in EMPTIES]
```
michamos: I think it would be more clear with
```
args_is_empty = all(arg in EMPTIES for arg in args)
```

inspire_schemas/utils.py
```diff
@@ -318,10 +318,11 @@ def filter_empty_parameters(func):
...
 
         if (
                 {'source', 'material'}.issuperset(my_kwargs) or not my_kwargs
-        ) and args == ():
+        ) and not my_args:
```
michamos: `args_is_empty`

PR #289, opened 2017-12-04
hep: relax length requirement on _files

jacquerie:


jacquerie:
> You need to set it to `0` instead.

Ok, done.

PR #292, opened 2017-12-07
elements/material: add various additional material

michamos:


jacquerie:
(Can you check if in `inspire-dojson` we're silently discarding materials that don't fit in the enum, instead of having them show up on Sentry?)

michamos:
we do, in https://github.com/inspirehep/inspire-dojson/blob/31b20aca6972b43e716af2dbfd805781652a7301/inspire_dojson/hep/rules/bd0xx.py#L104-L119 and https://github.com/inspirehep/inspire-dojson/blob/31b20aca6972b43e716af2dbfd805781652a7301/inspire_dojson/hep/rules/bd7xx.py#L79-L87.

PR #300, opened 2018-01-18
builders: reference: add more fields to publisher

szymonlopaciuk:
This adds support for `place` and `date` in `set_publisher`.

Signed-off-by: Szymon Łopaciuk <szymon.lopaciuk@cern.ch>

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -28,6 +28,7 @@ import re
 
 import six
 from inspire_utils.name import normalize_name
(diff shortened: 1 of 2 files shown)


michamos:
this is strange, as a `place` and a `date` are not part of a publisher. Two options: have a new `set_imprints` that can take all 3, or add two extra functions `set_publication_date` and `set_publication_place`.

szymonlopaciuk:
@michamos ah right, I don't know how I thought it made sense, on it

PR #303, opened 2018-01-18
utils: deduce year from volume for JHEP and JCAP

jacquerie:
Closes https://github.com/inspirehep/inspire-schemas/issues/302.

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -753,6 +753,8 @@ def convert_old_publication_info_to_new(publication_infos):
         journal_volume = _publication_info.get('journal_volume')
 
         if journal_title in _JOURNALS_WITH_YEAR_ADDED_TO_VOLUME and journal_volume and len(journal_volume) == 4:
+            was_last_century = int(journal_volume[:2]) > 50
+            _publication_info['year'] = int('19' + journal_volume[:2] if was_last_century else '20' + journal_volume[:2])
             _publication_info['journal_volume'] = journal_volume[2:]
             result.append(_publication_info)
             continue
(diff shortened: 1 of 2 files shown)


inspire_schemas/utils.py
```diff
@@ -753,6 +753,8 @@ def convert_old_publication_info_to_new(publication_infos):
         journal_volume = _publication_info.get('journal_volume')
 
         if journal_title in _JOURNALS_WITH_YEAR_ADDED_TO_VOLUME and journal_volume and len(journal_volume) == 4:
+            was_last_century = int(journal_volume[:2]) > 50
```
jacquerie: My apologies to the developer reading this in 2051.

michamos: you could use 90, JHEP and JCAP are from the nineties, but on the other hand, invenio 3 is meant to run 30 years, so 50 is probably enough ;)

tests/unit/test_utils.py
```diff
@@ -555,6 +555,34 @@ def test_convert_old_publication_info_to_new_handles_year_added_to_volumes():
...
 
+def test_convert_old_publication_info_to_new_deduces_year_from_volume():
+    schema = utils.load_schema('hep')
+    subschema = schema['properties']['publication_info']
+
+    publication_info = [
+        {
+            'artid': '065',
+            'journal_title': 'JHEP',
+            'journal_volume': '05',
```
michamos: doesn't seem to be testing what's declared.

jacquerie: It does seem that I'm dumb... Fixed.

PR #304, opened 2018-01-30
utils: fix get_license_from_url

jacquerie:
Adapt to ``isort@4.3.0``.

Make ``get_license_from_url`` handle CC public domain license URLs.

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -334,6 +334,11 @@ def test_get_license_from_url_handles_CC():
     assert utils.get_license_from_url(url) == 'CC BY-NC 4.0'
 
 
+def test_get_license_from_url_handles_CC_public_domain():
+    url = 'http://creativecommons.org/publicdomain/zero/1.0/'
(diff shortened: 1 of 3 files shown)


tests/unit/test_utils.py
```diff
@@ -334,6 +334,11 @@ def test_get_license_from_url_handles_CC():
     assert utils.get_license_from_url(url) == 'CC BY-NC 4.0'
 
 
+def test_get_license_from_url_handles_CC_public_domain():
+    url = 'http://creativecommons.org/publicdomain/zero/1.0/'
+    assert utils.get_license_from_url(url) == 'CC0 1.0'
```
jacquerie: @michamos is this what should happen here? It's not what it's happening, but at least it's no longer crashing...

michamos: :+1: that seems right

jacquerie: Now it's happening. Code doesn't look great, but it works...

PR #307, opened 2018-02-12
add successor option for related records

Glignos:
Signed-off-by: Glignos <glignos93@gmail.com>

### DIFF
--- a/inspire_schemas/records/elements/related_record.yml
+++ b/inspire_schemas/records/elements/related_record.yml
@@ -36,6 +36,27 @@ properties:
                   a project that finished and was superseded by the one in this
                   record (e.g. two successive runs of an experiment).
 
+            ``successor``
+                :MARC: ``510__w:b`` (for Institutions and Experiments),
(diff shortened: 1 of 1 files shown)


inspire_schemas/records/elements/related_record.yml
```diff
@@ -36,6 +36,27 @@ properties:
                   a project that finished and was superseded by the one in this
                   record (e.g. two successive runs of an experiment).
 
+            ``successor``
+                :MARC: ``510__w`` (for Institutions and Experiments),
```
michamos: `510__w:b`

inspire_schemas/records/elements/related_record.yml
```diff
@@ -36,6 +36,27 @@ properties:
...
+                    ``78502`` (for Journals) or the field comes from
```
michamos: `530__w:b`

Glignos: Ah yes indeed

PR #310, opened 2018-03-07
builders: normalize dates in LiteratureBuilder

vbalbp:

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -30,6 +30,8 @@ import warnings
 
 import idutils
 
+from inspire_utils.date import normalize_date
+
 from ..utils import (get_license_from_url,
                      normalize_collaboration,
                      validate,
@@ -272,15 +274,16 @@ class LiteratureBuilder(object):
         :param place: place of publication
         :type place: string
 
-        :param date: the date of publication
-        :type date: date
+        :param date: A (partial) date in any format.
+            The date should contain at least a year
+        :type date: string
 
         :rtype: dict
         """
 
         imprint = {}
         if date is not None:
-            imprint['date'] = date
+            imprint['date'] = normalize_date(date)
         if place is not None:
             imprint['place'] = place
         if publisher is not None:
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/literature.py
```diff
@@ -464,19 +466,21 @@ def add_publication_info(
     def add_imprint_date(self, imprint_date):
         """Add imprint date.
 
-        :type imprint_date: string. A formatted date is required (yyyy-mm-dd)
+        :type imprint_date: string. A date in any format including
```
michamos: this is not true, `month` and `day` are not required. What do you think of `A (partial) date in any format. The date should contain at least a year.` ?

inspire_schemas/builders/literature.py
```diff
@@ -464,19 +466,21 @@ def add_publication_info(
...
-            'date': imprint_date
+            'date': normalize_date(imprint_date)
         })
 
     @filter_empty_parameters
     def add_preprint_date(self, preprint_date):
         """Add preprint date.
 
-        :type preprint_date: string. A formatted date is required (yyyy-mm-dd)
+        :type preprint_date: string. A date in any format including
```
michamos: same comment

tests/unit/test_literature_builder.py
```diff
@@ -403,3 +403,13 @@ def test_publication_info_public_note():
     assert validate(result, subschema) is None
     assert expected == result
     assert 'publication_info' not in builder.record
+
+def test_preprint_date():
```
michamos: * please also add tests for the `add_imprint_date` and `add_book` (and add normalize_date there)
* the function name should be more explicit about what you are testing, so something like `test_preprint_date_normalizes_date`.

inspire_schemas/utils.py
```diff
@@ -34,13 +34,14 @@
...
 import six
 from idutils import is_orcid
-from inspire_utils.date import PartialDate
 from jsonschema import RefResolver, draft4_format_checker
 from jsonschema import validate as jsonschema_validate
 from pkg_resources import resource_filename
 from six.moves.urllib.parse import urlsplit
 from unidecode import unidecode
 
+from inspire_utils.date import PartialDate
```
michamos: this is wrong, probably a consequence of having `inspire_utils` locally as a module instead of installed in site-packages.

PR #311, opened 2018-03-08
utils: addition of ``draft`` validation

drjova:
* Adds ``validation_errors`` utils function, which returns a generator
  with all the schema errors. That is needed because normal JSONSchema
  validation returns only the first error.

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -35,7 +35,7 @@ import rfc3987
 import six
 from idutils import is_orcid
 from inspire_utils.date import PartialDate
-from jsonschema import RefResolver, draft4_format_checker
+from jsonschema import Draft4Validator, RefResolver, draft4_format_checker
 from jsonschema import validate as jsonschema_validate
 from pkg_resources import resource_filename
 from six.moves.urllib.parse import urlsplit
@@ -604,6 +604,32 @@ inspire_format_checker.checks('uri-reference', raises=ValueError)(
 inspire_format_checker.checks('orcid')(is_orcid)
 
 
+def _load_schema_for_record(data, schema=None):
+    """Load the schema from a given record.
+
+    Args:
+        data (dict): record data.
+        schema (Union[dict, str]): schema to validate against.
+
+    Returns:
+        dict: the loaded schema.
+
+    Raises:
+        SchemaNotFound: if the given schema was not found.
+        SchemaKeyNotFound: if ``schema`` is ``None`` and no ``$schema`` key was
+            found in ``data``.
+        jsonschema.SchemaError: if the schema is invalid.
+    """
+    if schema is None:
+        if '$schema' not in data:
+            raise SchemaKeyNotFound(data=data)
+        schema = data['$schema']
+
+    if isinstance(schema, six.string_types):
+        schema = load_schema(schema_name=schema)
+    return schema
+
+
 def validate(data, schema=None):
     """Validate the given dictionary against the given schema.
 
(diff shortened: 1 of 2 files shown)


inspire_schemas/utils.py
```diff
@@ -604,6 +604,33 @@ def load_schema(schema_name, resolved=False):
 inspire_format_checker.checks('orcid')(is_orcid)
 
 
+def _validate_schema(data, schema=None):
```
michamos: `_load_schema_for_record`?

inspire_schemas/utils.py
```diff
@@ -604,6 +604,33 @@ def load_schema(schema_name, resolved=False):
...
+    """Validate the schema.
```
michamos: no it doesn't

inspire_schemas/utils.py
```diff
@@ -638,6 +659,36 @@ def validate(data, schema=None):
     )
 
 
+def draft_validate(data, schema):
```
michamos: `validation_errors`? it doesn't really matter for the user that you did it by using the `Draft4Validator` directly instead of calling `validate`.

drjova:
@michamos done thanks :)

inspire_schemas/utils.py
```diff
@@ -638,6 +658,35 @@ def validate(data, schema=None):
     )
 
 
+def validation_errors(data, schema):
```
david-caro: Maybe rename to `get_validation_errors`? That will solve also the naming issues in https://github.com/inspirehep/inspire-next/pull/3239

inspire_schemas/utils.py
```diff
@@ -638,6 +658,35 @@ def validate(data, schema=None):
...
+    """Validation errors for a given record.
+
+    Args:
+        data (dict): record to validate.
+        schema (Union[dict, str]): schema to validate against. If it is a
+            string, it is intepreted as the name of the schema to load (e.g.
+            ``authors`` or ``jobs``). If it is ``None``, the schema is taken
+            from ``data['$schema']``. If it is a dictionary, it is used
+            directly.
```
david-caro: You might want to use `None` as the default value as in the `_load_schema_for_record` one.

inspire_schemas/utils.py
```diff
@@ -638,6 +658,35 @@ def validate(data, schema=None):
...
+        jsonschema.SchemaError: if the schema is invalid.
+    """
+    schema = _load_schema_for_record(data, schema)
+
+    errors = Draft4Validator(
+        schema,
+        resolver=LocalRefResolver.from_schema(schema),
+        format_checker=inspire_format_checker
+    )
+    return errors.iter_errors(data)
```
david-caro: Why not returning a list? If you prefer the iterator, maybe add a hint to the function name? `get_validation_errors_iter`

michamos: I think we should get used to iterators, it's the default for many things in Python 3 to which we will have to migrate sooner or later. For the same reason, adding `iter` seems superfluous to me: if it's plural, it's probably some collection you can iterate over, so either a list or an iterator, and the exact type is documented in the docstring.

About the `get`, I am opposed to prepending every function that doesn't mutate its arguments with `get`, and reserve it for functions that solely read things and perform no additional logic (see https://softwareengineering.stackexchange.com/a/198170 for a similar opinion). If you really want to have it in an as an imperative, maybe `find_validation_errors` though it doesn't add much info to `validation_errors`.

drjova: 👍 for ``validation_error``

PR #312, opened 2018-03-23
builders: fix urls in ReferenceBuilder

vbalbp:

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -33,7 +33,7 @@ from isbn import ISBN
 
 import idutils
 
-from ..utils import convert_old_publication_info_to_new, split_pubnote, fix_url
+from ..utils import convert_old_publication_info_to_new, split_pubnote, fix_reference_url
 
 
 # Matches any separators for author enumerations.
@@ -207,17 +207,14 @@ class ReferenceBuilder(object):
 
     def add_url(self, url):
         try:
-            self._add_uid(url, False)
+            self._add_uid(url, True)
         except ValueError:
             self._add_url(url)
 
     def _add_url(self, url):
-        fixed_url = fix_url(url)
+        fixed_url = fix_reference_url(url)
         self._ensure_reference_field('urls', [])
-        if fixed_url is not None:
-            self.obj['reference']['urls'].append({'value': fixed_url})
-        else:
-            self.obj['reference']['urls'].append({'value': url})
+        self.obj['reference']['urls'].append({'value': fixed_url})
 
     def add_refextract_authors_str(self, authors_str):
         """Parses individual authors from refextracted authors string."""
@@ -280,7 +277,7 @@ class ReferenceBuilder(object):
         except ValueError:
             self.add_misc(uid)
 
-    def _add_uid(self, uid, check_handle=True):
+    def _add_uid(self, uid, skip_handle=False):
         """Add unique identifier in correct field.
 
         The ``check_handle`` flag is used when adding a uid through the add_url function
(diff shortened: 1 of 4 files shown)


inspire_schemas/builders/references.py
```diff
@@ -274,40 +274,26 @@ def add_report_number(self, repno):
...
     def add_uid(self, uid):
         try:
             self._add_uid(uid)
-        except Exception:
+        except ValueError:
             self.add_misc(uid)
 
-    def _add_uid(self, uid):
-        """Add unique identifier in correct field."""
+    def _add_uid(self, uid, check_handle=True):
```
michamos: can you do it the other way around with `skip_handle=False` by default?

inspire_schemas/builders/references.py
```diff
@@ -206,9 +206,9 @@ def set_year(self, year):
             self.obj['reference']['publication_info']['year'] = year
 
     def add_url(self, url):
-        if self._is_uid(url):
-            self._add_uid(url)
-        else:
+        try:
+            self._add_uid(url, False)
```
michamos: `skip_handle=True`, which makes the meaning of the flag clearer (this is a good idea for flags in general).

inspire_schemas/utils.py
```diff
@@ -857,3 +857,51 @@ def _needs_a_hidden_pubnote(journal_title, journal_volume):
         result.append(_publication_info)
 
     return result
+
+
+def contains_bars_instead_of_slashes(string):
+    """Checking if the error of replacing ``/`` with ``|`` is present"""
```
michamos: * You could have a function `fix_url_bars_instead_of_slashes` that checks if it needs to do anything, and does the replacement if needed. Then in `fix_url`, you basically chain all the functions without `if`s.
* This should follow https://www.python.org/dev/peps/pep-0257/

inspire_schemas/utils.py
```diff
@@ -857,3 +857,51 @@ def _needs_a_hidden_pubnote(journal_title, journal_volume):
...
+    return 'http://' + string
+
+
+def replace_tilde(string):
+    """Replace unicode characters by their working equivalent"""
+    string = string.replace('\u223c', '~')
+    return string.replace('\u02dc', '~')
+
+
+def fix_url(url):
```
michamos: `fix_reference_url`

tests/unit/test_utils.py
```diff
@@ -802,3 +802,27 @@ def function_no_empty_args(arg1, arg2, source=None, material=None):
         assert False
 
     function_no_empty_args(arg1, arg2, source=source, material=material)
+
+
+def test_fix_url_add_http():
+    url_string = 'www.google.com/'
```
michamos: FYI, a neutral website you can use for this is `www.example.com`.

inspire_schemas/utils.py
```diff
@@ -857,3 +857,51 @@ def _needs_a_hidden_pubnote(journal_title, journal_volume):
...
+    """Return a url in a correct format.
+
+    Used to parse an incorect url to try to fix it
+    with the most common ocurrences for errors.
+    If the fixed url is still incorrect, it returns ``None``.
```
michamos: can you document the return value in the Google Napoleon style we are using (examples [here](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)).

inspire_schemas/utils.py
```diff
@@ -857,3 +857,51 @@ def _needs_a_hidden_pubnote(journal_title, journal_volume):
...
+    if is_http_missing(url):
+        url = add_http(url)
+
+    url = replace_tilde(url)
+
+    try:
+        rfc3987.parse(url, rule="URI")
+        return url
+    except ValueError:
+        return None
```
michamos: You could return the original url here instead of `None`, so you don't have to test the return value in the builder.

PR #313, opened 2018-03-27
elements: add external_system_identifiers to reference json schema

vbalbp:

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -49,6 +49,16 @@ RE_ARXIV_POST_2007_CLASS = re.compile(
     flags=re.I
 )
 
+# Matches CDS urls for id extraction
+CDS_MATCHER = re.compile(
+    r'^(https?://)?cds(web)?\.cern\.ch/record/(\d*)',
+    flags=re.I)
+
+# Matches ADS urls for id extraction
+ADS_MATCHER = re.compile(
+    r'^(https?://)?adsabs\.harvard\.edu/abs/(.*)',
+    flags=re.I)
+
 
 def _split_refextract_authors_str(authors_str):
     """Extract author names out of refextract authors output."""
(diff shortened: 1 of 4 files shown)


inspire_schemas/records/elements/reference.yml
```diff
@@ -268,5 +268,38 @@ properties:
         minItems: 1
         type: array
         uniqueItems: true
+external_system_identifiers:
```
jacquerie: Indentation doesn't look right: this should probably be nested one level deeper.

inspire_schemas/records/elements/reference.yml
```diff
@@ -268,5 +268,38 @@ properties:
         minItems: 1
         type: array
         uniqueItems: true
+    external_system_identifiers:
```
michamos: did you run `scripts/normalize_json.py`? you also need to generate examples and add them to the test fixtures for hep records.

inspire_schemas/builders/references.py
```diff
@@ -314,6 +326,26 @@ def _add_uid(self, uid, skip_handle=False):
             except Exception:
                 raise ValueError('Unrecognized uid type')
 
+    def is_CDS_id(self, uid):
+        """Check if ``uid`` corresponds to a CDS id"""
+        cds_matcher = re.compile(r'^.*cds\.cern\.ch/record/(\d*)\?.*')
```
michamos: * why the question mark? if you use the same pattern both to check and extract, you should define it only once.
* there are more domains for CDS, like `cdsweb`. I can try to find a list somewhere.

inspire_schemas/builders/references.py
```diff
@@ -314,6 +326,26 @@ def _add_uid(self, uid, skip_handle=False):
             except Exception:
                 raise ValueError('Unrecognized uid type')
 
+    def is_CDS_id(self, uid):
```
michamos: this should be a `staticmethod`.

inspire_schemas/builders/references.py
```diff
@@ -49,6 +49,16 @@
     flags=re.I
 )
 
+# Matches CDS urls for id extraction
+CDS_MATCHER = re.compile(
+    r'^.*cds\.cern\.ch/record/(\d*)\??.*',
```
michamos: * you should be more strict: `(https?//)?` instead of `.*` at the beginning.
* `cds(web)?` (cdsweb.cern.ch is valid too).
* why the end with the question mark? you don't have `$` at the end of the regexp and you are only matching digits in the group, so you don't care about what appears there.

inspire_schemas/builders/references.py
```diff
@@ -127,6 +137,26 @@ def _normalize_arxiv(obj):
         return '.'.join(m.group(2, 3))
 
 
+def is_CDS_id(uid):
```
michamos: `is_cds_url`

PR #314, opened 2018-04-04
elements: fix documentation reference json schema

vbalbp:

### DIFF
--- a/inspire_schemas/records/elements/reference.yml
+++ b/inspire_schemas/records/elements/reference.yml
@@ -77,14 +77,14 @@ properties:
         uniqueItems: true
     external_system_identifiers:
         description: |-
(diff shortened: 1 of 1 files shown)


inspire_schemas/records/elements/reference.yml
```diff
@@ -77,14 +77,14 @@ properties:
...
-            :MARC: ``035``
+            :MARC: ``999C5f``
             List of identifiers of this document on external systems.
         items:
             additionalProperties: false
             properties:
                 schema:
                     description: |-
-                        :MARC: ``035__9``
+                        :MARC: ``999C5f`` (part before ':')
```
michamos: use double-backticks instead of `'`

PR #320, opened 2018-05-22
builders: create an author builder

jacquerie:
Will be needed to create authors after disambiguation.

### DIFF
--- a/inspire_schemas/builders/authors.py
+++ b/inspire_schemas/builders/authors.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/authors.py
```diff
@@ -0,0 +1,44 @@
...
+
+class AuthorBuilder(object):
+    def __init__(self, author=None):
+        if author is None:
+            author = {}
+        self.obj = author
+
+    @filter_empty_parameters
+    def set_name(self, value):
+        self._ensure_field('name', {'value': normalize_name(value)})
```
michamos: `_ensure_field` does nothing if the field is already there as far as I understand. That's not what I would expect, as it means the builder is useless to change that field.

jacquerie: You're right, but then this wrong pattern has to be fixed all over the place in the signature builder, for example: https://github.com/inspirehep/inspire-schemas/blob/963237a736db12324136760c06479d41d6abed9c/inspire_schemas/builders/signatures.py#L99-L101

tests/unit/test_author_builder.py
```diff
@@ -0,0 +1,83 @@
...
+    assert validate(result, subschema) is None
+    assert expected == result
+
+
+def test_author_builder_set_name_can_be_called_multiple_times():
+    schema = load_schema('authors')
+    subschema = schema['properties']['name']
+
+    author = AuthorBuilder()
+    author.set_name('Richard Tower')
```
michamos: :smile:

PR #321, opened 2018-05-30
builders: fix _is_arxiv regex bug

vbalbp:

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -33,7 +33,7 @@ from isbn import ISBN
 
 import idutils
 
-from ..utils import convert_old_publication_info_to_new, split_pubnote, fix_reference_url
+from ..utils import convert_old_publication_info_to_new, split_pubnote, fix_reference_url, is_arxiv, normalize_arxiv
 
 
 # Matches any separators for author enumerations.
@@ -42,12 +42,6 @@ RE_SPLIT_AUTH = re.compile(r',?\s+and\s|,?\s*&|,|et al\.?|\(?eds?\.\)?',
 # Matches any stream of initials (A. B C D. -E F).
 RE_INITIALS_ONLY = re.compile(r'^\s*-?[A-Z]((\.|\s)\s*-?[A-Z])*\.?\s*$',
                               re.U)
-# Matches new style arXiv ID, with an old-style class specification
-# (Malformed, but appears in APS records)
-RE_ARXIV_POST_2007_CLASS = re.compile(
-    "(arxiv:)?(?:[a-z\-]+)(?:\.[a-z]{2})?/(\d{4})\.(\d{4,5})(v\d+)?$",
-    flags=re.I
-)
 
 # Matches CDS urls for id extraction
 CDS_MATCHER = re.compile(
@@ -101,40 +95,6 @@ def _split_refextract_authors_str(authors_str):
     ]
     res = [r for r in res if all(not f(r) for f in filters)]
     return res
-
-
-def _is_arxiv(obj):
-    """Return ``True`` if ``obj`` contains an arXiv identifier.
-
-    The ``idutils`` library only handles arXiv identifiers, e.g. strings
-    of the form ``arXiv:yymm.xxxxx``, but we sometimes have to deal with
-    arXiv references, which might contain more information separated by
-    a space. Therefore this helper wraps ``idutils`` to support this case.
-    """
-    arxiv_test = obj.split()
-    if not arxiv_test:
-        return False
-    return idutils.is_arxiv(arxiv_test[0]) \
-        or RE_ARXIV_POST_2007_CLASS.match(arxiv_test[0])
-
-
-def _normalize_arxiv(obj):
-    """Return a normalized arXiv identfier.
-
-    As in ``_is_arxiv``, we need to handle arXiv references as well
-    as arXiv identifiers. We also need to return a simpler arXiv
-    identifier than what ``idutils`` would output, so we use some
-    of its helpers instead of ``normalize_arxiv``.
-    """
-    obj = obj.split()[0]
-
-    m = idutils.is_arxiv_pre_2007(obj)
-    if m:
-        return ''.join(m.group(2, 4, 5))
-
-    m = idutils.is_arxiv_post_2007(obj) or RE_ARXIV_POST_2007_CLASS.match(obj)
-    if m:
-        return '.'.join(m.group(2, 3))
 
 
 def is_cds_url(uid):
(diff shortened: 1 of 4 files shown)


inspire_schemas/builders/references.py
```diff
@@ -45,7 +45,17 @@
 # Matches new style arXiv ID, with an old-style class specification
 # (Malformed, but appears in APS records)
 RE_ARXIV_POST_2007_CLASS = re.compile(
-    "(arxiv:)?(?:[a-z\-]+)(?:\.[a-z]{2})?/(\d{4})\.(\d{4,5})(v\d+)?$",
+    "(arxiv:)?([a-z\-]+)(?:\.[a-z]{2})?/(\d{4})\.(\d{4,5})(v\d+)?$",
```
michamos: I don't think this is correct, you should match the whole textual part for the category.
It's also easier to use identifiers. Like
```
"(arxiv:)?((?P<category>(?:[a-z\-]+)(?:\.[a-z]{2})?)/)?(P<identifier>\d{4})\.(\d{4,5})(v\d+)?$"

File: inspire_schemas/builders/references.py [hunk 2]
```diff
@@ -45,7 +45,17 @@
...
+    flags=re.I
+)
+
+RE_ARXIV_PRE_2007_CLASS = re.compile(
+    "(arxiv:)?([a-z\-]+)(\.[a-z]{2})?(/\d{4})(\d+)(v\d+)?$",
+    flags=re.I
+)
+
+RE_ARXIV_POST_2007 = re.compile(
```
michamos: You can remove this, as the other pattern will also match this.

inspire_schemas/builders/references.py
```diff
@@ -45,7 +45,17 @@
 # Matches new style arXiv ID, with an old-style class specification
 # (Malformed, but appears in APS records)
 RE_ARXIV_POST_2007_CLASS = re.compile(
-    "(arxiv:)?(?:[a-z\-]+)(?:\.[a-z]{2})?/(\d{4})\.(\d{4,5})(v\d+)?$",
+    "(arxiv:)?([a-z\-]+)(?:\.[a-z]{2})?/(\d{4})\.(\d{4,5})(v\d+)?$",
+    flags=re.I
+)
+
+RE_ARXIV_PRE_2007_CLASS = re.compile(
```
michamos: same idea.

inspire_schemas/builders/references.py
```diff
@@ -114,9 +124,20 @@ def _is_arxiv(obj):
...
+    matched_non_class = RE_ARXIV_POST_2007.match(arxiv_test[0])
+
+    if matched_non_class:
+        return True
+
+    matched_arxiv = RE_ARXIV_PRE_2007_CLASS.match(arxiv_test[0]) \
         or RE_ARXIV_POST_2007_CLASS.match(arxiv_test[0])
 
+    if matched_arxiv:
+        return matched_arxiv.group(2) in valid_arxiv_categories()
```
michamos: you should rather do something like this to account for casing and variations: https://github.com/inspirehep/inspire-schemas/blob/8ad58e4f18f42049f089f959dc7b8f2eda73c87a/inspire_schemas/utils.py#L402-L405

inspire_schemas/utils.py
```diff
@@ -957,3 +969,41 @@ def normalize_isbn(isbn):
...
+    modified here to work with two regular expressions instead
+    of three and adding a check for valid arxiv categories only"""
+    arxiv_test = obj.split()
+    if not arxiv_test:
+        return False
+
+    matched_arxiv = (RE_ARXIV_PRE_2007_CLASS.match(arxiv_test[0]) or
+                     RE_ARXIV_POST_2007_CLASS.match(arxiv_test[0]))
+
+    if matched_arxiv:
```
michamos: it's clearer to do it the other way around:
```
if not matched_arxiv:
    return False

# do the rest
```

inspire_schemas/utils.py
```diff
@@ -957,3 +969,41 @@ def normalize_isbn(isbn):
...
+        valid_arxiv_categories_lower = [category.lower() for category in valid_arxiv_categories()]
```
michamos: you only need this if you have matched a category, so this can go inside the `if`.

tests/unit/test_utils.py
```diff
@@ -76,35 +76,35 @@ def test_classify_field_ignores_case():
     assert expected == result
 
 
-def test_normalize_arxiv_category_returns_input_for_correct_category():
+def testnormalize_arxiv_category_returns_input_for_correct_category():
```
michamos: why did you remove the `_` after `test` here and in the following tests? too eager search & replace I guess.

inspire_schemas/utils.py
```diff
@@ -310,6 +310,18 @@
...
+RE_ARXIV_POST_2007_CLASS = re.compile(
+    "(arxiv:)?((?P<category>(?:[a-z\-]+)(?:\.[a-z]{2})?)/)?(?P<identifier>\d{4}\.\d{4,5})(v\d+)?$",
+    flags=re.I
+)
+
+RE_ARXIV_PRE_2007_CLASS = re.compile(
+    "(arxiv:)?(?P<category>(?P<extraidentifier>[a-z\-]+)(?:\.[a-z]{2})?)/(?P<identifier>\d{4}\d+)(v\d+)?$",
+    flags=re.I
+)
+
```
michamos: Can you put this above, with the other regexes?

inspire_schemas/utils.py
```diff
@@ -957,3 +969,43 @@ def normalize_isbn(isbn):
...
+    arxiv_test = obj.split()
+    if not arxiv_test:
+        return False
+
+    matched_arxiv = (RE_ARXIV_PRE_2007_CLASS.match(arxiv_test[0]) or
+                     RE_ARXIV_POST_2007_CLASS.match(arxiv_test[0]))
+
+    if not matched_arxiv:
+        return False
+    else:
```
michamos: the whole point of this is to be able to avoid the `else`, to remove one indentation level.

tests/unit/test_utils.py
```diff
@@ -850,3 +850,131 @@ def test_fix_reference_url_substitute_tilde():
...
+
+
+def test_normalize_arxiv_handles_new_identifiers_with_cls_prefix_and_ver():
+    expected = '1501.00001'
+    result = utils.normalize_arxiv('arXiv:hep-th.GT/1501.00001v1')
+
+    assert expected == result
+
+
+def text_normalize_arxiv_handles_uppercase():
```
michamos: * `test` instead of `text`
* can you add similar tests for `is_arxiv`, also with negative examples?

tests/unit/test_utils.py
```diff
@@ -850,3 +850,131 @@ def test_fix_reference_url_substitute_tilde():
...
+    expected = 'math/0312059'
+    result = utils.normalize_arxiv('MATH/0312059')
+
+    assert expected == result
+
+
+def text_normalize_arxiv_handles_uppercase_with_cls():
```
michamos: same

PR #323, opened 2018-06-07
builders: don't accept invalid ADS/CDS urls in references

vbalbp:


michamos:
can you also make the commit message more meaningful: `builders: don't accept invalid ADS/CDS urls in references`. In general, commit messages should reflect the high-level goal of the change, not the thing you actually did to the code.

PR #324, opened 2018-07-19
utils: handle unicode journals in build_pubnote

michamos:


inspire_schemas/utils.py
```diff
@@ -506,19 +506,22 @@ def split_pubnote(pubnote_str):
...
+    if title and volume and artid and artid != page_start:
+        pubnote_format = u'{title},{volume},{artid}'
+    elif title and volume and page_start and page_end:
+        pubnote_format = u'{title},{volume},{page_start}-{page_end}'
+    elif title and volume and page_start:
+        pubnote_format = u'{title},{volume},{page_start}'
+    elif title and volume:
+        pubnote_format = u'{title},{volume}'
+    else:
+        return None
```
david-caro: Is it ok to return `None` sometimes? Feels like mixing types xd

michamos: yes, this is used by `inspire_dojson` exclusively, and fields with `None` values are cleaned up at the end.

PR #329, opened 2018-09-21
builders: addition of ``acquisition source`` in authors.

drjova:

### DIFF
--- a/inspire_schemas/builders/authors.py
+++ b/inspire_schemas/builders/authors.py
@@ -33,16 +33,25 @@ from ..utils import EMPTIES, filter_empty_parameters
 
 class AuthorBuilder(object):
     """Author record builder."""
-    def __init__(self, author=None):
+    def __init__(self, author=None, source=None):
         if author is None:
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/authors.py
```diff
@@ -301,3 +312,56 @@ def add_private_note(self, note, source=None):
...
+        :type internal_uid: string
+
+        :param datetime: UTC datetime in ISO 8601 format
+        :type datetime: string
+        """
+        if date is not None:
+            if datetime is not None:
+                raise ValueError("Conflicting args: 'date' and 'datetime'")
+            warnings.warn("Use 'datetime', not 'date'", DeprecationWarning)
+            datetime = date
```
michamos: this is new, so no need to support both `date` and `datetime`, I think `datetime` is enough.

drjova:
@michamos updated!

drjova:
@michamos if I merge it'll not affect the versioning, right?

PR #332, opened 2018-10-17
utils: accept letters mid-volume in pub_info conversion

vbalbp:
When converting journals from Legacy to Labs,
```
'journal_title': 'Phys.Rev.',
'journal_volume': 'D43',
```
turns into
```
'journal_title': 'Phys.Rev.D',
'journal_volume': '43',
```
.
Previously,
```
'journal_title': 'Eur.Phys.J.', 
'journal_volume': 'A28S1',
```
would turn into
```
'journal_title': 'Eur.Phys.J.A', 
'journal_volume': '28',
```
with the fix made in this PR, it would turn into
```
'journal_title': 'Eur.Phys.J.A', 
'journal_volume': '28S1',
```
as it's expected.


vbalbp:
requests needs urllib3 version 1.23 and not 1.24, pinned it in setup.py so that dependencies are correct

PR #333, opened 2018-10-18
setup.py: fix urllib3 dependencies

vbalbp:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -180,8 +180,8 @@ def do_setup():
             'pyyaml',
             'rfc3987',
             'six',
-            # requests requires a urllib3 version of 1.23, we pin it down here to solve dependency problems
-            'urllib3==1.23',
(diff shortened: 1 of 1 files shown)


michamos:
This should be bumped in inspire-next instead https://github.com/requests/requests/blob/2c6a8426aebd853966747f2c851f551c583cb21a/setup.py#L53.

setup.py
```diff
@@ -180,8 +180,8 @@ def do_setup():
             'pyyaml',
             'rfc3987',
             'six',
-            # requests requires a urllib3 version of 1.23, we pin it down here to solve dependency problems
-            'urllib3==1.23',
+            # requests requires a urllib3 version of 1.23 or lower, we pin it down here to solve dependency problems
```
drjova:
```suggestion
            # requests requires a urllib3 version lower than 1.23, we pin it down here to solve dependency problems
```

PR #337, opened 2018-11-12
institutions: add GRID to external system identifiers

vbalbp:

### DIFF
--- a/inspire_schemas/records/institutions.yml
+++ b/inspire_schemas/records/institutions.yml
@@ -103,6 +103,26 @@ properties:
                 properties:
                     schema:
                         description: |-
+                            :MARC: ``035__9:GRID``
+                        enum:
+                        - GRID
+                        minLength: 1
+                        type: string
+                    value:
+                        description: |-
+                            :MARC: ``035__a``
+                        minLength: 1
+                        pattern: ^grid\.\d+\.\w+$
+                        type: string
+                required:
+                - schema
+                - value
+                title: GRID identifier
+                type: object
+            -   additionalProperties: false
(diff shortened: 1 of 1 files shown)


vbalbp:
Based on examples found on INSPIRE:
```
1. | grid.268323.e
2. | grid.1008.9
3. | grid.38142.3c
4. | grid.170205.1
5. | grid.412151.2
6. | grid.451133.1
7. | grid.57686.3a
8. | grid.440662.2
```
I think this ``^grid\.\d+\.\d\w?$`` makes sense as a regular expression for the GRID identifier.

michamos:
apparently the last part is hexadecimal: look at the regex in https://www.wikidata.org/wiki/Property:P2427. I could not find any reliable source for that regex, so maybe relax it to `^grid\.\d+\.\w+$`.

inspire_schemas/records/institutions.yml
```diff
@@ -119,6 +119,26 @@ properties:
...
+                    value:
+                        description: |-
+                            :MARC: ``035__a``
+                        minLength: 1
+                        pattern: ^grid\.\d+\.\d\w?$
+                        type: string
+                required:
+                - schema
+                - value
+                title: HAL identifier
```
michamos: wrong title

vbalbp:
[On the same webpage that you linked](https://www.wikidata.org/wiki/Property:P2427), there is a section called ``format as a regular expression`` which uses this expression: ``grid\.\d{4,6}\.[0-9a-f]{1,2}``.
I think it would make sense to use it instead?

PR #342, opened 2018-12-10
builders: fix author position sorting

harunurhan:
* Fixes the sorting when there are positions with and without start_date

* Fixes the sorting when there are positions with and without rank

### DIFF
--- a/inspire_schemas/builders/authors.py
+++ b/inspire_schemas/builders/authors.py
@@ -34,7 +34,9 @@ from ..utils import EMPTIES, filter_empty_parameters, load_schema
 
 
 RANKS = load_schema('elements/rank')['enum']
+RANKS.append(None)
 INSTITUTION_RANK_TO_PRIORITY = {rank: -idx for (idx, rank) in enumerate(RANKS)}
+EARLIEST_DATE = PartialDate.loads('1000')
 
 
 class AuthorBuilder(object):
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/authors.py
```diff
@@ -34,7 +34,9 @@
 
 
 RANKS = load_schema('elements/rank')['enum']
+RANKS.append(None)
 INSTITUTION_RANK_TO_PRIORITY = {rank: -idx for (idx, rank) in enumerate(RANKS)}
+EARLIEST_DATE = PartialDate.parse('1-1-1000')
```
michamos: `.loads('1000-01-01')` so you don't need to go through the parser.

harunurhan:
@michamos adressed!

inspire_schemas/builders/authors.py
```diff
@@ -34,7 +34,9 @@
 
 
 RANKS = load_schema('elements/rank')['enum']
+RANKS.append(None)
```
michamos: why is this needed?

harunurhan: if there is one with rank and one without, it fails. But it should be `any rank > no rank`.

harunurhan: maybe better to put a comment for this.

PR #345, opened 2019-01-07
builder: literature ignore invalid ``dois``

drjova:


inspire_schemas/builders/literature.py
```diff
@@ -189,7 +189,12 @@ def add_doi(self, doi, source=None, material=None):
         """
         if doi is None:
             return
-        doi = idutils.normalize_doi(doi)
+
+        try:
+            doi = idutils.normalize_doi(doi)
+        except AttributeError:
+            return None
```
michamos:
```suggestion
            return
```
the function is not returning any value, but mutating `self.obj`.

drjova:
has been updated

PR #347, opened 2019-01-09
records: add institutions ids to hep

vbalbp:


inspire_schemas/records/hep.yml
```diff
@@ -280,6 +280,35 @@ properties:
                     minItems: 1
                     type: array
                     uniqueItems: true
+                affiliations_ids:
```
michamos: could you make this `affiliation_identifiers`? please also add it to the integration test fixture.

PR #353, opened 2019-02-04
Added support for files and filenames field in files/documents/figures

pazembrz:


inspire_schemas/builders/literature.py
```diff
@@ -32,12 +32,14 @@
...
-                     filter_empty_parameters,
-                     EMPTIES)
+from ..utils import (
+    get_license_from_url,
+    normalize_collaboration,
+    normalize_isbn,
+    validate,
+    filter_empty_parameters,
+    EMPTIES
+)
```
michamos: 🙇 indentation master

inspire_schemas/builders/literature.py
```diff
@@ -844,31 +847,33 @@ def set_curated(self, curated=True):
...
+            file['url'] = kwargs['url']
         else:
             raise TypeError("Required argument 'url' should not be 'falsey'.")
+        return file
 
-        for dict_key in ('caption', 'label', 'material'):
-            if locals()[dict_key] is not None:
-                figure[dict_key] = locals()[dict_key]
+    @filter_empty_parameters
+    def add_figure(self, key, url, **kwargs):
```
michamos: The issue if you do it like this is that it's not clear from the signature which args the function should receive. I think it would be better to do it like previously (even if the `locals` thing is not very nice).

inspire_schemas/builders/literature.py
```diff
@@ -844,31 +847,33 @@ def set_curated(self, curated=True):
...
-    ):
-        """Add a figure."""
-        figure = self._sourced_dict(source)
-
-        if key:
-            figure['key'] = key
+    def _check_metadata_for_file(self, **kwargs):
+        file = self._sourced_dict(kwargs.get('source'))
+        if kwargs.get('key'):
+            file['key'] = kwargs['key']
```
michamos: do we want to allow providing a key? I thought this was supposed to be a hash of the contents.

inspire_schemas/records/hep.yml
```diff
@@ -759,7 +759,7 @@ properties:
                 material:
                     $ref: elements/material.json
                 original_url:
-                    format: uri
+                    format: string
```
michamos: why string?

pazembrz: Because url itself is also a string. This fields should be compatible.

drjova:
@pazembrz the tests are failling

PR #355, opened 2019-02-27
Pin json schema

monaawi:


michamos:
What's wrong with `jsonschema` 3.0? I don't think we should randomly pin packages without documenting why we do this, as this will cause lots of technical debt down the line.

monaawi:
It is causing tests to fail on inspire-query-parser. And the library is already pinned in inspire-next.

michamos:
This seems to be the issue: https://travis-ci.org/inspirehep/inspire-query-parser/jobs/499296043#L556.

PR #356, opened 2019-03-27
builder: add email address with hidden field

harunurhan:
* Adds or updates (if hidden is not None) with `hidden` field to
indicate if email is public or not.

* NSPIR-2157

### DIFF
--- a/inspire_schemas/builders/authors.py
+++ b/inspire_schemas/builders/authors.py
@@ -110,18 +110,28 @@ class AuthorBuilder(object):
         self.obj['name'].setdefault('native_names', []).append(name)
 
     @filter_empty_parameters
-    def add_email_address(self, email):
+    def add_email_address(self, email, hidden=None):
         """Add email address.
 
         Args:
-            :param email: public email of the author.
+            :param email: email of the author.
             :type email: string
+
+            :param hidden: if email is public or not.
+            :type hidden: boolean
         """
-        existing_emails = get_value(self.obj, 'email_addresses.value', [])
-        if email not in existing_emails:
-            self._append_to('email_addresses', {
-                "value": email
-            })
+        existing_emails = get_value(self.obj, 'email_addresses', [])
+        found_email = next(
+            (existing_email for existing_email in existing_emails if existing_email.get('value') == email),
+            None
+        )
+        if found_email is None:
+            new_email = {'value': email}
+            if hidden is not None:
+                new_email['hidden'] = hidden
+            self._append_to('email_addresses', new_email)
+        elif hidden is not None:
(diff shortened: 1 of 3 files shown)


inspire_schemas/builders/authors.py
```diff
@@ -110,18 +110,27 @@ def add_native_name(self, name):
...
+        found_email = next(
+            (existing_email for existing_email in existing_emails if existing_email.get('value') == email),
+            None)
+        if found_email is None:
+            new_email = {'value': email}
+            if hidden is not None:
+                new_email['hidden'] = hidden
+            self._append_to('email_addresses', new_email)
+        elif hidden is not None:
+            found_email['hidden'] = hidden
```
drjova: ```self._append_to('email_addresses', new_email)```?

inspire_schemas/builders/authors.py
```diff
@@ -110,18 +110,27 @@ def add_native_name(self, name):
...
         """
-        existing_emails = get_value(self.obj, 'email_addresses.value', [])
-        if email not in existing_emails:
-            self._append_to('email_addresses', {
-                "value": email
-            })
+        existing_emails = get_value(self.obj, 'email_addresses', [])
+        found_email = next(
+            (existing_email for existing_email in existing_emails if existing_email.get('value') == email),
+            None)
```
drjova: this is confusing to read, maybe with better indentation will be easier to follow:

```python
found_email = next(
    (existing_email for existing_email in existing_emails if existing_email.get('value') == email),
    None
)
```

drjova:
isort issues on travis

inspire_schemas/builders/authors.py
harunurhan: I would prefer a regular `for` loop but it's not "pythonic" 😄

harunurhan:
@drjova not caused by the PR :/, isort is supposed to fix this file.

drjova:
@harunurhan commit message typo ``INSPIR-2157``

inspire_schemas/builders/authors.py
drjova: you can always do it simple:

```
existing_emails = [existing_email for existing_email in existing_emails if existing_email.get('value') == email]

if existing_emails:
```

drjova:
indeed isort should ignore this file

drjova:
In the latest isort, released few days ago:

Fixed issue #909 - skip and skip-glob are not enforced when using settings-path.

inspire_schemas/builders/authors.py
harunurhan: this won't break the loop.

PR #359, opened 2019-05-10
literature: add record params

harunurhan:
Enables adding record objects for various fields:

* publication_info: conference_record, journal_record
* accelerator_experiment: record
* author: record

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -25,6 +25,7 @@
 """Literature builder class and related code."""
 
 from __future__ import absolute_import, division, print_function
+from six import string_types
 
 import warnings
 
@@ -228,7 +229,8 @@ class LiteratureBuilder(object):
                     source=None,
                     ids=(),
                     emails=(),
-                    alternative_names=()):
+                    alternative_names=(),
+                    record=None):
         """Make a subrecord representing an author.
 
         Args:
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/literature.py
```diff
@@ -222,6 +223,7 @@ def add_author(self, author):
 
     @filter_empty_parameters
     def make_author(self, full_name,
+                    record=None,
```
michamos: you should add it at the end, otherwise it breaks when using positional args, and update the docstring.

inspire_schemas/builders/literature.py
```diff
@@ -546,6 +554,19 @@ def add_accelerator_experiments_legacy_name(self, legacy_name):
             'legacy_name': legacy_name
         })
 
+    @filter_empty_parameters
+    def add_accelerator_experiment(self, legacy_name, record=None):
+        """Add legacy name in accelerator experiment.
+
+        :type legacy_name: string
```
michamos: add `record` to docstring.

tests/unit/test_literature_builder.py
```diff
@@ -478,6 +478,37 @@ def test_add_reference():
...
+    legacy_name = 'FNAL-E-0900'
+    experiment_record = {'$ref': 'http://url/api/experiments/123'}
+    builder.add_accelerator_experiment('FNAL-E-0900', record=experiment_record)
+    assert builder.record['accelerator_experiments'] == [{
+        'legacy_name': legacy_name,
+        'record': experiment_record
+    }]
+
+
+def test_publication_info_public_note():
```
michamos: wrong copy/paste?

inspire_schemas/builders/literature.py
harunurhan: but I hope nobody is calling it like that :)

PR #360, opened 2019-05-13
Jobs schema

michamos:
This adapts the job schema as discussed with Fermilab @hoc3426 @cleggm1.

### DIFF
--- a/inspire_schemas/records/jobs.yml
+++ b/inspire_schemas/records/jobs.yml
@@ -16,68 +16,152 @@ properties:
         type: array
         uniqueItems: true
     _private_notes:
+        description: |-
+            :MARC: ``595``
+
+            These notes are only visible to privileged users, not regular
+            users.
         items:
             $ref: elements/sourced_value.json
         minItems: 1
+        title: List of private notes
         type: array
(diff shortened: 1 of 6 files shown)


inspire_schemas/records/jobs.yml
```diff
@@ -16,68 +16,152 @@ properties:
...
             $ref: elements/json_reference.json
         minItems: 1
-        title: Deleted Records
         type: array
         uniqueItems: true
     description:
+        description: |-
+            :MARC: ``520__a``
+
+            This may contain rich text in Markdown format.
```
michamos: @drjova Not 100% sure this is the right decision vs a whitelist of HTML tags and attributes. If we migrate records, this will require sanitizing HTML + conversion to markdown anyway, and the `CKEditor 5` Markdown backend is still experimental, which might cause some issues.

michamos: After discussing IRL, we will stick with a subset HTML in order to avoid experimental tooling, parsing ambiguities and extra steps. HTML will be sanitized with [`bleach`](https://github.com/mozilla/bleach) in the jobs builder.

PR #362, opened 2019-06-06
builder: Add builder for Jobs

pazembrz:
* Added builder for Jobs

INSPIR-2430

### DIFF
--- a/inspire_schemas/builders/jobs.py
+++ b/inspire_schemas/builders/jobs.py
@@ -1,0 +1,351 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE-SCHEMAS.
+# Copyright (C) 2019 CERN.
+#
+# INSPIRE-SCHEMAS is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE-SCHEMAS is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE-SCHEMAS; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Jobs builder class and related code."""
+import logging
+import six
+
+from inspire_schemas.utils import validate, filter_empty_parameters, EMPTIES
+from inspire_utils.date import normalize_date
+
+
+LOG = logging.getLogger(__name__)
+
+
+class JobBuilder(object):
+    """Job record builder."""
+
+    def __init__(self, record=None, source=None,):
+        """Initialize builder with provided source and/or job record
+
+        Args:
+            source (string): sets the default value for the source fields of the curent
+            job record, which captures where the information that the builder populates
+            comes from
+            record (dict): sets the default value for the the current record, in order
+            to edit an already existent record
+        """
+        if record is None:
+            record = {
+                '_collections': ['Jobs'],
+                'status': 'pending'
+            }
+        self.record = record
+        self.source = source
+
+    def _ensure_field(self, field_name, default_value, obj=None):
+        if obj is None:
+            obj = self.record
+        if field_name not in obj:
+            obj.update({field_name: default_value})
+
+    def _ensure_list_field(self, field_name, default_value=None, obj=None):
+        if default_value is None:
+            default_value = []
+        self._ensure_field(field_name, default_value, obj)
+
+    def _ensure_dict_field(self, field_name, default_value=None, obj=None):
+        if default_value is None:
+            default_value = {}
+        self._ensure_field(field_name, default_value, obj)
+
+    def _sourced_dict(self, source=None, **kwargs):
+        """Like ``dict(**kwargs)``, but where the ``source`` key is special."""
+        if source:
+            kwargs['source'] = source
+        elif self.source:
+            kwargs['source'] = self.source
+        return kwargs
+
+    @filter_empty_parameters
+    def _append_to(self, field, element=None, default_list=None, **kwargs):
+        """Append the ``element`` to the ``field`` of the record.
+        This method is ``smart``: it does nothing if ``element`` is empty and
+            creates ``field`` if it does not exit yet.
+        Args:
+
+            field (string): the name of the field of the record to append to
+            element: the element to append
+            default_list (list): Default list which should be sert when field is missing
+        Kwargs: Arguments from which a dictionary will be built if element is empty
+        """
+        if default_list is None:
+            default_list = []
+        if element not in EMPTIES:
+            self._ensure_list_field(field, default_list)
+            if element not in self.record[field]:
+                self.record[field].append(element)
+        elif kwargs:
+            if 'record' in kwargs and isinstance(kwargs['record'], six.string_types):
+                kwargs['record'] = {'$ref': kwargs['record']}
+            self._ensure_list_field(field, default_list)
+            if kwargs not in self.record[field]:
+                self.record[field].append(kwargs)
+
+    def validate_record(self):
+        """Validate the record in according to the hep schema."""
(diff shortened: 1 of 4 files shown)


inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE-SCHEMAS.
+# Copyright (C) 2017 CERN.
```
drjova: 2019

tests/integration/test_builders.py
```diff
@@ -50,6 +60,11 @@ def input_data_hep():
     return load_file('input_data_hep.yaml')
 
 
+@pytest.fixture('module')
+def job_data():
+    return load_json_file('jobs_example.json')
+
```
drjova: why ```json``? in all other cases we have ``.yaml`` no?

pazembrz: There was already json in there, there were no yaml. I didn't want to loose time to convert etc. If you think that using yaml here give us any benefit I can do this.

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+        self.record = record
+        self.source = source
+
+    def _ensure_field(self, field_name, default_value, obj=None):
+        if obj is None:
+            obj = self.record
+        if field_name not in obj:
+            obj.update({field_name: default_value})
+
+    def _ensure_type(self, field_name, requested_type, default_value, obj=None):
```
michamos: what's the point of this? to me it seems like a poor man's `JSONschema` validation.

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+        Args:
+
+            field (string): the name of the field of the record to append to
+            element: the element to append
+            default_list (list): Default list which should be sert when field is missing
+        Kwargs: Arguments from which a dictionary will be built if element is empty
+        """
+        if default_list is None:
+            default_list = []
+        if element:
```
michamos: the other builders use `element in EMPTIES`: if a flag is set to `False` or a number to `0`, you don't want to drop it.

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+            accelerator (str)
+            curated_relation (bool)
+            experiment (str)
+            institution (str)
+            legacy_name (str)
+            record (dict): URL to the referenced resource.
+            When it's string, a new object will be created for record
+        """
+        if isinstance(record, six.string_types):
+            record = {'$ref': record}
```
michamos: this is not needed. Just choose what should be passed based on how it's done for `literature`.

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+        Kwargs: Arguments from which a dictionary will be built if element is empty
+        """
+        if default_list is None:
+            default_list = []
+        if element:
+            self._ensure_list_field(field, default_list)
+            if element not in self.record[field]:
+                self.record[field].append(element)
+        elif kwargs:
+            if 'record' in kwargs and isinstance(kwargs['record'], six.string_types):
```
michamos: you don't need to do this: choose one way and stick to it, look at how it's done in `literature`.

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+        """Add arxiv_category to ``arxiv_categories`` field
+
+        Args:
+            category (str)
+        """
+        if category is not 'other':
+            self._append_to('arxiv_categories', category)
+
+    @filter_empty_parameters
+    def add_field_of_research(self, research):
```
michamos: why duplicate this?

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+        """Add email or url to a list in proper subfield of ``reference_letters``.
+
+        Args:
+            reference_email (str): Email which will be added to ``emails`` list.
+            reference_url (dict): Url with description which will be added to ``urls``
+            list.
+        """
+        main_key = 'reference_letters'
+        email_key = 'emails'
+        url_key = 'urls'
```
michamos: why the parametrization here?

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+        """
+        entry = {
+            'value': value
+        }
+        if description:
+            entry['description'] = description
+        return entry
+
+    @filter_empty_parameters
+    def _add_reference(self, reference_email=None, reference_url=None):
```
michamos: what's the point of this function? you basically have a big `if` that checks which argument has been passed and does different things based on that, so you might as well split it in separate functions.

inspire_schemas/builders/jobs.py
```diff
@@ -0,0 +1,393 @@
...
+    def set_description(self, description):
+        """Set description of job
+
+        Args:
+            description (str): Job description
+        """
+        self.record['description'] = description
+
+    def set_status(self, status):
+        """Set status to the job. By default it sets it to PENDING.
```
michamos: It doesn't, it's just that `pending` is the default if this is not called.

inspire_schemas/builders/jobs.py
pazembrz: It's not validation at all. It's just easy way to create proper field with the type I want, and replace if it's different than I want.

michamos: not sure I understand the point. If the field has the wrong type, it logs an error and sets it to the default value. Why would you ever need to do that if the field didn't have the wrong type in the first place?

inspire_schemas/builders/jobs.py
pazembrz: What? this is helpful, I do not want to check if record is properly passed in every place.

inspire_schemas/builders/jobs.py
pazembrz: Because in form this is called completely different, so calling builder.add_arxiv_category(field_of_research)  is confusing.

inspire_schemas/builders/jobs.py
pazembrz: because this keys are used in more than in one place in this function. It's easier, and less error prone to keep them like that.

inspire_schemas/builders/jobs.py
pazembrz: because it still adding stuff in the same `field` called `reference_letters` and I wanted to stay consistent with this. Edit one key only from one place.

inspire_schemas/builders/jobs.py
pazembrz: Right, I changed function but I forgot to change comment.

inspire_schemas/builders/jobs.py
pazembrz: I'm not doing this if the field do not have the wrong type in the first place... Logging is only to inform that something is wrong. 
But probably you are right, I don't have to check this, as data should be correct.

inspire_schemas/builders/jobs.py
michamos: I still disagree here after discussing with you, as having a function with a big `if` is basically the same as having a function with a boolean parameter, which is an [anti-pattern](https://softwareengineering.stackexchange.com/a/148058).

inspire_schemas/builders/jobs.py
michamos: having two functions doing exactly the same thing is even more confusing I think, but whatever.

inspire_schemas/builders/jobs.py
pazembrz: We have much more anti patterns which are more serious than two, two lines ifs. And we are still adding worse anti-patterns everyday.
And as I said, this method makes code more contained in my opinion.

michamos: yes, I agree it's not worth spending more time arguing about, that's why I dismissed my review.

drjova: I don't understand 
>  And we are still adding worse anti-patterns everyday.

This is not how we work, we shouldn't add any anti-patterns. Even if we have, as you say, we shouldn't add more even if there are smaller or bigger. IMHO this not the right approach.

pazembrz: Sorry for that, I meant something different.

pazembrz: Still comments from others would be appreciated. As for now this discussion is only between my and Micha's convictions. I'm not saying that my opinion is the only truth and I am not saying that Micha is wrong, so second opinion is advised.

ammirate: I'll keep it short. I would simplify as much as possible, which means removing `_add_reference`, since it's not adding anything more than just having the logic in the other two functions and would require to add more tests.

PR #371, opened 2019-07-02
literature builder: add external_system_identifier

tsgit:
When harvesting OSTI via hepcrawl the `osti-id` of the record should be added to the literature record as `external_system_identifier`. This adds the necessary method to the LiteratureBuilder.

Signed-off-by: Thorsten Schwander <thorsten.schwander@gmail.com>

### DIFF
--- a/inspire_schemas/builders/literature.py
+++ b/inspire_schemas/builders/literature.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE-SCHEMAS.
-# Copyright (C) 2017 CERN.
+# Copyright (C) 2017, 2019 CERN.
 #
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/literature.py
```diff
@@ -691,6 +691,21 @@ def add_url(self, url):
...
+        :type extid: string
+
+        :param schema: identifies the external system for the given identifier
+        :type schema: string
+        """
+        self._append_to('external_system_identifiers', {
+            'schema': schema,
+            'value': extid,
+        })
+
```
michamos: looks OK. Could you also add a test?

tsgit: 4 tests added

PR #372, opened 2019-07-24
builders: fix unicode issues in LiteratureBuilder

michamos:


drjova:
@michamos do we have to commit all resolved json?

michamos:
whoops, my bad

drjova:
shall we add it to .gitignore as well?

michamos:
:+1: I'm already doing it

michamos:
actually those were leftovers from previous code, so not needed.

PR #375, opened 2019-08-14
setup: Update pinning to urllib3

pazembrz:


michamos:
@pazembrz is this still needed?

PR #376, opened 2019-08-27
schemas: Add missing minLength to '_bucket' in all schemas

pazembrz:


michamos:
already included in #379

PR #385, opened 2019-11-06
builders: refactor using inheritance

jacquerie:
Code duplication of helper methods such as `_ensure_field` led to subtly diverging implementations across the builders. Let's put a stop to this by making all record builders inherit from a single `RecordBuilder` class containing the implementation of these helper methods.


michamos:
Thanks @jacquerie!

PR #386, opened 2020-01-15
records: limit commas in name to two

miguelgrc:
* INSPIR-2974

### DIFF
--- a/tests/unit/test_author_builder.py
+++ b/tests/unit/test_author_builder.py
@@ -21,6 +21,8 @@
 # or submit itself to any jurisdiction.
 
 from __future__ import absolute_import, division, print_function
+import pytest
+import jsonschema
 
 from inspire_schemas.api import load_schema, validate
 from inspire_schemas.builders.authors import AuthorBuilder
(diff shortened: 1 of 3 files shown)


tests/unit/test_author_builder.py
```diff
@@ -917,10 +917,10 @@ def test_add_advisor_normalizes_name():
     subschema = schema['properties']['advisors']
 
     author = AuthorBuilder()
-    author.add_advisor('Riccardo Torres Jr')
+    author.add_advisor('Riccardo Torres')
 
     expected = [{
-        "name": 'Torres, Riccardo, Jr.',
```
michamos: As you can tell by this test, the commit is wrong. In most fields, two commas should be allowed.

miguelgrc: Then I think there's some issue in the [refinement](https://its.cern.ch/jira/browse/INSPIR-2974) @drjova

drjova: @michamos it should be up to two commas per name?

michamos: yes, precisely because of the suffixes like here.

drjova: 👍 @miguelgrc my bad sorry

miguelgrc: No worries @drjova, I'll change it :)

PR #387, opened 2020-02-17
ReferenceBuilder: handle additional pubnotes

michamos:
* INSPIR-3190

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -29,6 +29,7 @@ import re
 import six
 from inspire_utils.date import normalize_date
 from inspire_utils.name import normalize_name
+from inspire_utils.record import get_value
 from isbn import ISBN
 
 import idutils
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/references.py
```diff
@@ -355,3 +360,41 @@ def set_page_artid(self, page_start=None, page_end=None, artid=None):
...
+
+        Yields:
+            dict: a schema-compliant reference element for each additional
+            pubnote.
+        """
+        miscs = get_value(self.obj, 'reference.misc', [])
+        if not miscs:
+            return
+
+        for (index, misc) in enumerate(miscs[:]):
```
drjova: why ``miscs[:]`` and not ``miscs``?

michamos: because I want to mutate the array while iterating over it, which is not allowed in Python 3 (and also requires some arithmetic to update the index when deleting elements), so I make a (shallow) copy first.

inspire_schemas/builders/references.py
```diff
@@ -355,3 +360,41 @@ def set_page_artid(self, page_start=None, page_end=None, artid=None):
             publication_info['page_end'] = page_end
         if artid:
             publication_info['artid'] = artid
+
+    def pop_additional_pubnotes(self):
```
drjova:
```suggestion
    def remove_additional_pubnotes(self):
```

michamos: As discussed IRL, it doesn't just remove, but also yields them, somewhat like `pop` on standard python containers, so I think pop is more appropriate.

PR #391, opened 2020-04-16
utils: handle None journal_title in convert_publication_info_old_to_new

MJedr:


tests/unit/test_utils.py
```diff
@@ -647,6 +647,30 @@ def test_convert_old_publication_info_to_new_handles_year_added_to_volumes_when_
     assert expected == result
 
 
+def test_convert_old_publication_info_to_new_handles_year_added_to_volumes_when_journal_title_is_none():
```
michamos, line 650: here it's not `None`, it's not present, so
```suggestion
def test_convert_old_publication_info_to_new_handles_year_added_to_volumes_when_no_journal_title():
```

PR #392, opened 2020-04-29
seminars: add new schema type

michamos:
* INSPIR-3489
* Also updates `normalize_yaml.py` script to work on Python 3 instead of
  Python 2.

### DIFF
--- a/inspire_schemas/records/seminars.yml
+++ b/inspire_schemas/records/seminars.yml
@@ -1,0 +1,252 @@
+$schema: http://json-schema.org/schema#
+additionalProperties: false
+properties:
+    $schema:
+        format: uri
+        minLength: 1
+        type: string
+    _collections:
+        items:
+            enum:
+            - Seminars
+            minLength: 1
+            type: string
+        minItems: 1
+        title: Collections to which this record belongs
+        type: array
+        uniqueItems: true
+    _private_notes:
+        description: |-
+            These notes are only visible to privileged users, not regular
+            users.
+        items:
+            $ref: elements/sourced_value.json
+        minItems: 1
+        title: List of private notes
+        type: array
+        uniqueItems: true
+    abstract:
+        $ref: elements/sourced_value.json
+    acquisition_source:
+        $ref: elements/acquisition_source.json
+    address:
+        $ref: elements/address.json
+        minItems: 1
+        title: Location of the seminar
+        type: array
+        uniqueItems: true
+    contact_details:
+        items:
+            $ref: elements/contact.json
+        minItems: 1
+        title: Contact information
+        type: array
+        uniqueItems: true
+    control_number:
+        description: |-
+            Read-only field.
+        title: ID of current record
+        type: integer
+    core:
+        title: Whether this record is CORE
+        type: boolean
+    deleted:
+        title: Whether this record has been deleted
+        type: boolean
+    deleted_records:
+        description: |-
+            List of records that were deleted because they were replaced by
+            this one. This typically happens when merging two records: one of
+            them gets enriched with the information of the other one, which is
+            then superfluous and gets deleted.
+
+            For the opposite concept, see :ref:`new_record`.
+        items:
+            $ref: elements/json_reference.json
+        minItems: 1
+        type: array
+        uniqueItems: true
+    end_datetime:
+        format: date-time
+        minLength: 1
+        title: Seminar end datetime in UTC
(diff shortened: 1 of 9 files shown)


inspire_schemas/records/seminars.yml
```diff
@@ -0,0 +1,263 @@
+$schema: http://json-schema.org/schema#
+additionalProperties: false
+properties:
+    $schema:
+        format: uri
+        minLength: 1
+        type: string
+    _bucket:
```
monaawi, line 8: is this still needed?

inspire_schemas/records/seminars.yml
```diff
@@ -0,0 +1,263 @@
...
+        minItems: 1
+        type: array
+        uniqueItems: true
+required:
+- _collections
+- end_datetime
+- start_datetime
+- timezone
+- speakers
+- title
```
monaawi, line 261: in the mockup, ```inspire_categories``` is also mandatory in the submission form. Should we make it required here as well?

inspire_schemas/records/seminars.yml
```diff
@@ -0,0 +1,263 @@
...
+        items:
+            enum:
+            - Seminars
+            minLength: 1
+            type: string
+        minItems: 1
+        title: Collections to which this record belongs
+        type: array
+        uniqueItems: true
+    _private_notes:
```
monaawi, line 29: are these the ```additional information``` in the form? Or does ```additional information``` correspond to ```public_notes```? If yes, then how do these get populated?

inspire_schemas/records/seminars.yml
```diff
@@ -0,0 +1,263 @@
+$schema: http://json-schema.org/schema#
+additionalProperties: false
+properties:
```
monaawi, line 3: In the form we also have the field ```institution```, I also saw there was a discussion about it on zulip. But I see you didn't add the ```institution``` property, do we not want it?

tests/unit/test_api.py
```diff
@@ -167,3 +167,28 @@ def test_validate_raises_on_invalid_orcid():
...
+        '$schema': 'http://json-schema.org/schema#',
+        'type': 'string',
+        'format': 'timezone',
+    }
+
+    utils.validate(data, schema)
+
+
+def test_validate_raises_on_invalid_timezone():
+    data = "SevenKingdoms/KingsLanding"
```
monaawi, line 185: :laughing:

inspire_schemas/records/seminars.yml
michamos: additional information maps to public notes, private notes are private stuff only visible to curators (probably not needed here).

inspire_schemas/records/seminars.yml
michamos: :+1: makes sense

inspire_schemas/records/seminars.yml
michamos: :shrug: it's still there for all other record types, so I added it too.

inspire_schemas/records/seminars.yml
michamos: we are dropping `institution` in the end: https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/topic/seminars/near/195704729

inspire_schemas/records/seminars.yml
monaawi: we don't have ```invenio_records_files``` anymore, I think we can safely remove it from here. I don't know why we have them in other record types.

monaawi: I would still remove this as we definitely don't need it

inspire_schemas/records/seminars.yml
monaawi: We should either remove these if we don't need them or add a way of adding/updating them if we think they are useful

michamos: they're there in all collections, even jobs, so I think it's better to keep it even though there's no easy way to edit them currently

inspire_schemas/records/seminars.yml
michamos: done

inspire_schemas/records/seminars.yml
monaawi: ok, you're right. Let's keep them then.

PR #393, opened 2020-05-04
seminars: add builder

harunurhan:
INSPIR-3493

### DIFF
--- a/inspire_schemas/builders/seminars.py
+++ b/inspire_schemas/builders/seminars.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE-SCHEMAS.
-# Copyright (C) 2019 CERN.
+# Copyright (C) 2020 CERN.
 #
 # INSPIRE-SCHEMAS is free software; you can redistribute it
 # and/or modify it under the terms of the GNU General Public License as
@@ -28,6 +28,7 @@ from __future__ import absolute_import, division, print_function
 
 import six
 
+from inspire_utils.name import normalize_name
 from inspire_schemas.builders.builder import RecordBuilder
 from inspire_schemas.utils import (
     filter_empty_parameters,
@@ -37,10 +38,10 @@ from inspire_schemas.utils import (
 from inspire_utils.date import normalize_date
 
 
-class ConferenceBuilder(RecordBuilder):
-    """Conference record builder."""
+class SeminarBuilder(RecordBuilder):
+    """Seminar record builder."""
 
-    _collections = ['Conferences']
+    _collections = ['Seminars']
 
     @staticmethod
     def _prepare_url(value, description=None):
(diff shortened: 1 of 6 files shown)


inspire_schemas/builders/seminars.py
```diff
@@ -0,0 +1,275 @@
...
+            latitude (float): latitude of the location.
+            longitude (float): longitude of the location.
+            place_name (str): name of the specific place where this is located.
+            postal_address (str): full postal address in original language.
+            postal_code (str): postal code of the location.
+            state (str): state or province of the location.
+        """
+        address = {}
+
+        if cities:
```
harunurhan, line 89: do we have a better way of doing this in place already? @michamos

michamos: I don't think so

inspire_schemas/builders/seminars.py
```diff
@@ -0,0 +1,275 @@
...
+    def add_series(self, name, number=None):
+        """Add conference series.
+
+        Args:
+            name (str): name of the conference series.
+            number (int): number of the conference series.
+        """
+        serie_object = self._sourced_dict(name=name)
+        if number:
+            serie_object['number'] = number
```
michamos, line 210: I think `_sourced_dict` already filters empty values.

inspire_schemas/builders/seminars.py
```diff
@@ -0,0 +1,275 @@
...
+            title (str): title for the current document.
+            subtitle (str): subtitle for the current document.
+            source (str): source for the given title.
+        """
+        title = self._sourced_dict(
+            source,
+            title=title,
+        )
+        if subtitle is not None:
+            title['subtitle'] = subtitle
```
michamos, line 227: same

inspire_schemas/builders/seminars.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        Args:
+            name (str): name of the contact.
+            record (dict): dictionary with ``$ref`` pointing to proper record.
+            If string, then will be converted to proper dict.
+            affiliations (list): list of affiliations objects
+        """
+        self._append_to(
+            'speakers',
+            name=name,
```
michamos, line 138: you should either use the signature builder here or manually call `normalize_name`.

inspire_schemas/builders/seminars.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def set_end_datetime(self, date=None):
+        """
+        Args:
+            date (str)
+        """
+        if date is not None:
+            self.record['end_datetime'] = date
+
+    def set_start_datetime(self, date=None):
```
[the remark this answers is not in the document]
harunurhan: I didn't want var name to overlap with native `datetime` module.

inspire_schemas/builders/seminars.py
harunurhan: I copy pasted this from `conferences` actually, but I can't see how it filters:

```python
    def _sourced_dict(self, source=None, **kwargs):
        if source:
            kwargs['source'] = source
        elif self.source:
            kwargs['source'] = self.source
        return kwargs
```

michamos: sorry, I was confused with `_append_to`. Maybe it makes sense doing the check there?

inspire_schemas/builders/seminars.py
```diff
@@ -0,0 +1,276 @@
...
+        """
+        Args:
+            name (str): name of the contact.
+            record (dict): dictionary with ``$ref`` pointing to proper record.
+            If string, then will be converted to proper dict.
+            affiliations (list): list of affiliations objects
+        """
+        self._append_to(
+            'speakers',
+            name=normalize_name(name),
```
michamos, line 139: can you add a test for name normalization?

harunurhan: it's in the unit tests.

inspire_schemas/builders/seminars.py
harunurhan: personally I don't like implicit logic in the functions, but sure I can put it there

PR #397, opened 2020-05-27
global: addition of countries

drjova:
* Adds countries form iso_3166-1 and iso_3166-3.

* INSPIR-3451


inspire_schemas/api.py
```diff
@@ -26,5 +26,63 @@
...
+        return country_code_to_country
+
+    @cached_property
+    def name_to_code(self):
+        country_name_to_code = {}
+        for country in self.countries:
+            name = self.get_country_name(country)
+            code = self.get_country_code(country)
+            country_name_to_code[name] = code
+        return country_name_to_code
```
michamos, line 88: why make this thing so complex? To me it looks like a useless optimization to do this lazily. It's not like the list is huge in the first place, and we have plenty of this kind of static lists already (arXiv categories, mapping of arXiv categories to INSPIRE categories, etc.). I think we need exactly two functions: `country_code_to_name` and `country_name_to_code`, which can do a simple lookup in a couple of dicts that are built at module import time.

drjova: So what you're suggesting is to load json files and build dicts on module import time? 

For me I still don't see any value loading and keeping these iso files, imo we don't care about compatibility with iso we just need a map which we can easily update, which probably will not be that often.

michamos: the point is easily update. How will you easily update it if you don't know what has changed? look at the diff and apply it manually?

inspire_schemas/records/elements/country_code.yml
```diff
@@ -1,264 +1,285 @@
...
-    Netherlands Antilles, NT - Neutral Zone)
+description: 'Country code according to `ISO 3166-1 alpha-2
+
+  <https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2>`_ with a few additions
+
+  (CS - Czechoslovakia, YU - Yugoslavia, AB - Serbia and Montenegro, AN -
+
+  Netherlands Antilles, NT - Neutral Zone)'
 enum:
-- AB
```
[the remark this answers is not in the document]
drjova: I will remove this file, it's generated automatically with the union.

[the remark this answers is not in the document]
drjova: I've updated item, still is less readable and dirty imo. I don't understand why design patterns are considered complex.

setup.py
```diff
@@ -121,11 +121,43 @@ def _find(basepath, extension='.yml'):
...
+
+def _generate_country_code(base_path):
+    import yaml
+    from inspire_schemas.utils import COUNTRY_CODE_TO_NAME
+
+    countries_code_path = os.path.join(base_path, 'elements', 'country_code.yml')
+    data = {
+        'maxLength': 2,
+        'minLength': 2,
+        'title': 'ISO 3166-1 alpha 2 country code',
```
michamos, line 143:
```suggestion
        'title': 'ISO 3166-1 or 3166-3 alpha 2 country code',
```

tests/unit/test_utils.py
```diff
@@ -1282,3 +1282,27 @@ def test_sanitize_html():
...
+@pytest.mark.parametrize("country_code,expected", [("GR", "Greece"), ("MC", "Monaco")])
+def test_code_to_name(country_code, expected):
+    assert utils.country_code_to_name(country_code) == expected
+
+
+@pytest.mark.parametrize(
+    "country_code,expected", [("TW", "Taiwan"), ("VE", "Venezuela")]
+)
+def test_code_to_common_name(country_code, expected):
+    assert utils.country_code_to_name(country_code) == expected
```
michamos, line 1308: maybe add tests for the 3166-3 case to make sure they're part of the list?

setup.py
```diff
@@ -121,11 +121,43 @@ def _find(basepath, extension='.yml'):
...
+def _generate_country_file():
+    import json
+    from inspire_schemas.utils import COUNTRY_NAME_TO_CODE
+
+    schemas_js_dir = os.path.join(
+        os.path.dirname(__file__),
+        'schemas_js'
+    )
+    with open(os.path.join(schemas_js_dir, 'src', 'countries.json'), 'w+') as json_fd:
+        json.dump(list(COUNTRY_NAME_TO_CODE.keys()), json_fd)
```
michamos, line 133: don't you want to serialize the json object you'll need directly, to avoid doing it in the JS code?
```suggestion
        json.dump([{"value": name} for name in COUNTRY_NAME_TO_CODE]), json_fd)
```

inspire_schemas/countries/iso_3166-1.json
```diff
@@ -0,0 +1,1677 @@
+{
```
michamos, line 1: can you add somewhere a link to the origin of these files, so we don't need to go hunt for them later?

setup.py
drjova: Good idea!

inspire_schemas/countries/iso_3166-1.json
drjova: I have added a comment in the utils.

drjova:
@michamos updated

package.json
```diff
@@ -1,6 +0,0 @@
-{
-    "dependencies": {
-        "json-schema-faker": "^0.4.1",
-        "moment": "^2.19.1"
-    }
-}
```
michamos, line 6: actually this is needed for the script to generate schema examples

drjova: 👍

drjova: wait this changing a lot actually, is what I was avoiding to do. I thought it was a mistake If I take advantage of this I can use the default travis deployment for npm

.npmignore
```diff
@@ -0,0 +1,17 @@
+inspire_schemas/
+tests/
+dist/
+.vscode/
+.pytest_cache/
+.eggs/
+inspire_schemas.egg-info/
+docs/
+scripts/
+
```
harunurhan, line 10: I think you need `!js/dist` because AFAIR `.gitignore` is merged into `.npmignore`

rollup.config.js
```diff
@@ -0,0 +1,10 @@
+import json from '@rollup/plugin-json';
+
+export default {
+  input: 'js/index.js',
+  output: {
+    file: 'build/index.js',
+    format: 'cjs'
```
harunurhan, line 7: I think common js might work already but we can also build for `es` too.

This would require to add `"module": "dist/index.es.js"` to `package.json` and

```js
    {
      file: pkg.module,
      format: 'es',
      sourcemap: true
    }
```

in the output config.

rollup.config.js
```diff
@@ -0,0 +1,10 @@
+import json from '@rollup/plugin-json';
+
+export default {
+  input: 'js/index.js',
+  output: {
+    file: 'build/index.js',
```
harunurhan, line 6: you can `import pkg from './package.json'` and `pkg.main` here.

setup.py
```diff
@@ -121,11 +121,43 @@ def _find(basepath, extension='.yml'):
...
+def _generate_country_file():
+    import json
+    from inspire_schemas.utils import COUNTRY_NAME_TO_CODE
+
+    schemas_js_dir = os.path.join(
+        os.path.dirname(__file__),
+        'js'
+    )
+    with open(os.path.join(schemas_js_dir, 'countries.json'), 'w+') as json_fd:
+        json.dump([{"value": name} for name in COUNTRY_NAME_TO_CODE.keys()], json_fd)
```
harunurhan, line 133: `{"value": name}` logic is just part of our ui and how select field works. So I wouldn't include it here.

So if it was only needed for inspirehep/ui, this could be a list of countries BUT the plan is also use this for record-editor: https://github.com/inspirehep/record-editor/blob/709315867ef84d69e5270bbb029d942634717e39/src/app/shared/constants/iso-639.ts#L1

Where we need the actual map. So I think It's best to just export the map `countryCodeToName` since it is the superior one then we can create `{ value: name }` list out of it for inspirehep/ui

drjova:
@harunurhan `offical_name` is not part of the US.

harunurhan:
That became part of the our logic in the `backend`, after the US was created probably.

drjova:
Then I'll close this PR, it's already the third time I implemented this due to refinement issues.

harunurhan:
I don't think it's refinement issue, and you reimplemented it 3 times, and it even requires reimplementation, and closing the PR is the right way here.

All needed is not to make sure not to break the logic you're replacing, which is

```py
def country_name_to_code(country_name):
    try:
        country = pycountry.countries.get(name=country_name)
    except KeyError:
        country = pycountry.countries.get(official_name=country_name)
    except KeyError:
        country = pycountry.countries.get(common_name=country_name)

    return country.alpha_2 if country else None
```

apparently it's only used from `name` to `code` conversion but not the other way.. I don't know why is that like that.. perhaps there is not even a change required but as I said, just need to make it sure.

setup.py
drjova: shall we prepare it already or do it in the ``inspirehep/ui``

michamos:
Why can't you add simply add the `official_name` thing here?

drjova:
@harunurhan @michamos do you mean as a different dict? The only problem I see this way is what we'll export to the UI.

michamos:
actually I don't think the `official_name` thing is needed if we use the same name everywhere. All countries have a `name`, so `official_name` doesn't seem necessary.

drjova:
At the moment we have this logic:
https://github.com/inspirehep/inspire-schemas/pull/397/files#diff-1d0970a763dbbbe297382c68c9c55ac8R346-R347

michamos:
That's ok, as I said, all countries have a `name`, which is more human-friendly than `official_name`.

drjova:
ok so I'll leave it this way

drjova:
@harunurhan thank you for your suggestions, I have updated the PR.

.npmignore
```diff
@@ -0,0 +1 @@
+!js/dist
```
harunurhan, line 1: well now it needs to ingore all the rest though like `inspire_schemas/` etc, so that we don't push python files etc.

I think you can use `npm pack` to see what actually would be published on npm.

js/index.js
```diff
@@ -0,0 +1,5 @@
+import * as countries from './countries.json';
```
harunurhan, line 1: should we still call this `countries`? maybe something like `countryCodeToName` would be better since we might need to export a map `countryNameToCode` at some point.

package.json
```diff
@@ -1,6 +1,24 @@
...
+    "scripts": {
+        "test": "echo \"Error: no test specified\" && exit 1",
+        "build": "rollup --config"
+    },
+    "author": "Inspire",
+    "license": "MIT",
+    "publishConfig": {
+        "access": "public"
+    },
     "dependencies": {
```
harunurhan, line 16: these should be also `devDependencies`

drjova: I am not sure, whatever use these packages run it with ``--dev`` that's why I left it.

PR #399, opened 2020-06-03
seminars: add related records field

monaawi:
* INSPIR-3571

### DIFF
--- a/inspire_schemas/builders/seminars.py
+++ b/inspire_schemas/builders/seminars.py
@@ -144,6 +144,20 @@ class SeminarBuilder(RecordBuilder):
         )
 
     @filter_empty_parameters
+    def add_literature_record(self, record, curated_relation=None):
+        """
+        Args:
+            record (dict): dictionary with ``$ref`` pointing to proper record.
+            If string, then will be converted to proper dict.
+            curated_relation (bool): mark if relation is curated [NOT REQUIRED]
+        """
+        self._append_to(
+            'literature_records',
+            record=record,
+            curated_relation=curated_relation,
+        )
+
+    @filter_empty_parameters
     def add_join_url(self, value, description=None):
         """Add url dict to ``urls`` list.
 
--- a/inspire_schemas/records/seminars.yml
+++ b/inspire_schemas/records/seminars.yml
@@ -219,6 +219,21 @@ properties:
         title: List of speakers
         type: array
         uniqueItems: true
+    literature_records:
+        items:
+            additionalProperties: false
+            properties:
+                curated_relation:
+                    type: boolean
+                record:
+                    $ref: elements/json_reference.json
+            required:
+                - record
+            type: object
+        minItems: 1
+        title: List of literature records related to the seminar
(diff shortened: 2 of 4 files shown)


inspire_schemas/builders/seminars.py
```diff
@@ -143,6 +143,20 @@ def add_speaker(self, name=None, record=None, affiliations=None, ids=None, curat
             curated_relation=curated_relation,
         )
 
+    @filter_empty_parameters
+    def add_paper_record(self, record=None, curated_relation=None):
```
drjova, line 147: shouldn't be this ``literature_record`` or something similar? ``paper_record`` seems odd.

monaawi: that's the name I had in the US, I find it a bit weird too, should I change it?

drjova: I'm not sure why, @michamos do you know why we name it ``paper_record``?

monaawi: I would even call it ```related_records``` since this is the name of the field on the UI

inspire_schemas/records/seminars.yml
```diff
@@ -219,6 +219,21 @@ properties:
         title: List of speakers
         type: array
         uniqueItems: true
+    related_records:
```
michamos, line 222: sorry, but this name is confusing, as `related_records` mean something else entirely in all other collections (see https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/elements/related_record.yml). I think `literature_records` is a good name though.

inspire_schemas/records/seminars.yml
```diff
@@ -219,6 +219,21 @@ properties:
...
+            properties:
+                curated_relation:
+                    type: boolean
+                record:
+                    $ref: elements/json_reference.json
+            required:
+                - record
+            type: object
+        minItems: 1
+        title: List of related records
```
michamos, line 234:
```suggestion
        title: List of literature records related to the seminar
```

inspire_schemas/builders/seminars.py
```diff
@@ -143,6 +143,20 @@ def add_speaker(self, name=None, record=None, affiliations=None, ids=None, curat
             curated_relation=curated_relation,
         )
 
+    @filter_empty_parameters
+    def add_literature_record(self, record=None, curated_relation=None):
```
michamos, line 147: I think `record` should not have a default value, to signal it is required.

tests/unit/test_seminar_builder.py
```diff
@@ -506,3 +506,25 @@ def test_add_speaker():
...
+            'record': {'$ref': 'http://literature/another'},
+            'curated_relation': True
+        },
+    ]
+
+    builder = SeminarBuilder()
+    builder.add_literature_record(
+        record={'$ref': 'http://literature/1'}
+    )
+    builder.add_literaturerecord(
```
michamos, line 526: typo

PR #400, opened 2020-06-04
npm: move dependencies to dev

drjova:


.travis.yml
```diff
@@ -50,3 +50,9 @@ deploy:
     on:
       branch: master
       python: '2.7'
+  - provider: script
```
michamos, line 53: this is a duplicate I think

drjova: I want to run this on ``tags`` for the npm package

drjova: https://docs.travis-ci.com/user/deployment-v2/conditional#tag

harunurhan: I don't understand why we would need this.

drjova: to avoid having versions like ``61.3.0-2-g3e91dc0``

PR #401, opened 2020-06-09
experiments: add inspire_classification instead of class number

MJedr:
* INSPIR-3603

### DIFF
--- a/inspire_schemas/records/experiments.yml
+++ b/inspire_schemas/records/experiments.yml
@@ -245,17 +245,13 @@ properties:
         uniqueItems: true
     inspire_classification:
         items:
+            $ref: elements/inspire_experiment_classification.json
             description: |-
                 :MARC: ``372__a``
 
-                Category to which this experiment belongs. The possible values
-                can be found on the `experiment categories
-                page <http://inspirehep.net/info/Experiments/list>`_ on legacy.
+                Category to which this experiment belongs.
 
-                :example: ``3.1.1`` for ``ultra-short baseline, reactor, neutrino
-                    (flavor) experiment``
             minLength: 1
-            pattern: \d+(\.\d+){0,2}
             type: string
         minItems: 1
         type: array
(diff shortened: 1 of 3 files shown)


inspire_schemas/records/elements/inspire_classification.yml
```diff
@@ -0,0 +1,70 @@
+description: |-
+  Category to which this experiment belongs.
+  The possible values can be found on the
+   `experiment categories page <http://inspirehep.net/info/Experiments/list>`_ on legacy.
```
michamos, line 4: You can remove that, as this will become the primary source once we migrate experiments to labs.

inspire_schemas/records/elements/inspire_classification.yml
```diff
@@ -0,0 +1,70 @@
+description: |-
+  Category to which this experiment belongs.
```
michamos, line 2:
```suggestion
  An INSPIRE experiment category. This is a hierarchical classification, with `|` used as a separator.
```

michamos:
In theory this is an incompatible change, so the major version should be bumped, but this should affect only `inspire-dojson` (for experiments) and `inspirehep` itself (where we lock packages), so I think it's ok if we bump only the minor version (with `sem-ver: new feature` in the commit message), otherwise we'll need to bump a lot of other packages (hepcrawl, query-parser, matcher, merger, etc.). WDYT @drjova ?

drjova:
@michamos sounds good

inspire_schemas/records/experiments.yml
```diff
@@ -245,17 +245,14 @@ properties:
...
         items:
+            $ref: elements/inspire_classification.json
             description: |-
                 :MARC: ``372__a``
 
-                Category to which this experiment belongs. The possible values
-                can be found on the `experiment categories
-                page <http://inspirehep.net/info/Experiments/list>`_ on legacy.
+                An INSPIRE experiment category.
+                This is a hierarchical classification, with `|` used as a separator.
```
michamos, line 253:
```suggestion
                Category to which this experiment belongs.
```

inspire_schemas/records/experiments.yml
```diff
@@ -245,17 +245,14 @@ properties:
         uniqueItems: true
     inspire_classification:
         items:
+            $ref: elements/inspire_classification.json
```
michamos, line 248: can you rename this file to `inspire_experiment_classification`?

PR #403, opened 2020-06-25
seminars: add accessibility info

miguelgrc:
* INSPIR-3566

### DIFF
--- a/inspire_schemas/records/seminars.yml
+++ b/inspire_schemas/records/seminars.yml
@@ -32,6 +32,9 @@ properties:
     address:
         $ref: elements/address.json
         title: Location of the seminar
+    captioned:
+        title: Whether this seminar contains captions
+        type: boolean
     contact_details:
         items:
             $ref: elements/contact.json
@@ -130,6 +133,12 @@ properties:
             Contains a reference to the record replacing the current one, if it
             is marked as :ref:`deleted`.
         title: Record replacing this one
+    material_urls:
+        items:
+            $ref: elements/url.json
+        title: Links to related material such as slides.
+        type: array 
+        uniqueItems: true               
     public_notes:
         description: |-
             Any notes about the document that do not fit into another field.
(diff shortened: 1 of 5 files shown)


inspire_schemas/records/seminars.yml
```diff
@@ -32,6 +32,9 @@ properties:
     address:
         $ref: elements/address.json
         title: Location of the seminar
+    captions:
```
michamos, line 35: we use adjectives for this kind of stuff if possible:
```suggestion
    captioned:
```

inspire_schemas/records/seminars.yml
```diff
@@ -130,6 +133,12 @@ properties:
             Contains a reference to the record replacing the current one, if it
             is marked as :ref:`deleted`.
         title: Record replacing this one
+    material:
```
michamos, line 136: why not
```suggestion
    material_urls:
```
to be consistent with `join_urls`?

inspire_schemas/records/seminars.yml
```diff
@@ -130,6 +133,12 @@ properties:
...
+        items:
+            $ref: elements/url.json
+        title: Links to PDFs or slides.
```
michamos, line 139: Slides can be PDFs but don't have to:
```suggestion
        title: Links to related materials such as slides.
```
To me it's not clear if we want only slides here or also potentially other stuff.

inspire_schemas/records/seminars.yml
```diff
@@ -130,6 +133,12 @@ properties:
             Contains a reference to the record replacing the current one, if it
             is marked as :ref:`deleted`.
         title: Record replacing this one
+    material_urls:
+        items:
+            $ref: elements/url.json
+        title: Links to related material such as slides.
+        type: array 
+        uniqueItems: true
```
michamos, line 141: don't forget to add this to the integration fixture: https://github.com/inspirehep/inspire-schemas/blob/master/tests/integration/fixtures/seminars_example.json

miguelgrc: I did!

PR #404, opened 2020-07-23
utils: fix conversion of journal title and volume to new version

MJedr:
Ref: #1132

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -877,8 +877,6 @@ def convert_old_publication_info_to_new(publication_infos):
         try:
             journal_title = _JOURNALS_RENAMED_OLD_TO_NEW[journal_title]
             _publication_info['journal_title'] = journal_title
-            result.append(_publication_info)
-            continue
         except KeyError:
             pass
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_utils.py
```diff
@@ -831,9 +840,12 @@ def test_convert_old_publication_info_to_new_does_not_double_letters():
         },
     ]
     result = utils.convert_old_publication_info_to_new(publication_info)
+    result_2 = utils.convert_old_publication_info_to_new(publication_info_2)
 
     assert utils.validate(result, subschema) is None
+    assert utils.validate(result_2, subschema) is None
     assert expected == result
+    assert expected == result_2
```
michamos, line 848: can you create a new test case for this instead?

inspire_schemas/utils.py
```diff
@@ -915,7 +914,7 @@ def convert_old_publication_info_to_new(publication_infos):
         hidden = _publication_info.pop('hidden', None)
         if hidden:
             hidden_publication_infos.append(_publication_info)
-        else:
+        elif _publication_info not in result:
```
michamos, line 917: can't you remove the `result.append` on L880 instead? I think it would be more clear.

PR #409, opened 2020-10-27
literature: stricter file url validation

harunurhan:
`format: uri-reference` accepts pretty much anything.
but we want file urls (documents.url and figure.url)
to be either absolute url or relative `/api/files` url.

### DIFF
--- a/inspire_schemas/records/hep.yml
+++ b/inspire_schemas/records/hep.yml
@@ -812,7 +812,8 @@ properties:
                         form is ``/files/bucket/key``. It can temprorarily be
                         the url to download the document from, until actually
                         downloaded.
-                    format: uri-reference
+                    format: 'uri-references'
+                    pattern: ^(https?://|/api/files/).*
                     minLength: 1
                     type: string
             required:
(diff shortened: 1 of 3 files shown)


inspire_schemas/records/hep.yml
```diff
@@ -812,7 +812,7 @@ properties:
                         form is ``/files/bucket/key``. It can temprorarily be
                         the url to download the document from, until actually
                         downloaded.
-                    format: uri-reference
+                    pattern: ^(https?://|/api/files/).*
```
michamos, line 815: I think we want both, it has to be a `uri-reference`, that also matches this pattern

inspire_schemas/records/hep.yml
```diff
@@ -972,7 +972,7 @@ properties:
 
                         Relative URL to the file containing the figure. Its
                         form is ``/files/bucket/key``.
-                    format: uri-reference
+                    pattern: ^(https?://|/api/files/).*
```
michamos, line 975: same

michamos (approved):

PR #410, opened 2020-11-03
authors: add hidden to advisors, positions, projects

MJedr:
Ref: inspirehep/inspirehep#1347

### DIFF
--- a/inspire_schemas/builders/authors.py
+++ b/inspire_schemas/builders/authors.py
@@ -226,7 +226,17 @@ class AuthorBuilder(RecordBuilder):
         self._append_to('arxiv_categories', category)
 
     @filter_empty_parameters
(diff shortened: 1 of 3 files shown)


inspire_schemas/builders/authors.py
```diff
@@ -226,7 +226,7 @@ def add_arxiv_category(self, category):
         self._append_to('arxiv_categories', category)
 
     @filter_empty_parameters
-    def add_institution(self, institution, start_date=None, end_date=None, rank=None, record=None, curated=False, current=False):
+    def add_institution(self, institution, start_date=None, end_date=None, rank=None, record=None, curated=False, current=False, hidden=None):
```
drjova, line 229: why not `hidden=False`? since it should be bool and it's how we're using it for other params for example `curated`

PR #414, opened 2020-11-24
json_reference: create separate schema for each record type

MJedr:


inspire_schemas/utils.py
```diff
@@ -1079,3 +1080,47 @@ def sanitize_html(text):
...
+            for field in reference_fields
+            if not re.search("deleted_records|new_record|self", field)
+        ]
+        for reference_field in reference_fields_to_process:
+            index_name = get_value(schema, reference_field).split(" ")[0]
+            reference_search_path = re.sub(
+                "properties.|items.|.description$", "", reference_field
+            )
+            if index_name in references:
+                references[index_name].append((schema_name, reference_search_path))
```
michamos, line 1125: you should also add `related_records.$ref` here (which doesn't have a name specified, so you need to hardcode it) .

inspire_schemas/utils.py
```diff
@@ -1079,3 +1080,47 @@ def sanitize_html(text):
...
+def get_paths(schema):
+    for key, val in schema.items():
+        if isinstance(val, dict):
+            for subkey in get_paths(val):
+                yield "{key}.{subkey}".format(key=key, subkey=subkey)
+        else:
+            yield key
+
+
+def get_references_for_schemas():
```
michamos, line 1094:
```suggestion
def get_refs_to_schemas():
```
Please also add a docstring. Also memoization would be good, as we'll call this often and it will never change during the run of the app.

inspire_schemas/utils.py
```diff
@@ -1079,3 +1080,47 @@ def sanitize_html(text):
...
+        "experiments",
+        "institutions",
+        "conferences",
+        "seminars",
+        "jobs",
+        "journals",
+    ]
+    references = {}
+    for schema in schemas:
+        references[schema] = []
```
michamos, line 1107:
```suggestion
    references = defaultdict(list)
```

inspire_schemas/utils.py
```diff
@@ -1079,3 +1080,47 @@ def sanitize_html(text):
...
+            field
+            for field in reference_fields
+            if not re.search("deleted_records|new_record|self", field)
+        ]
+        for reference_field in reference_fields_to_process:
+            index_name = get_value(schema, reference_field).split(" ")[0]
+            reference_search_path = re.sub(
+                "properties.|items.|.description$", "", reference_field
+            )
+            if index_name in references:
```
michamos, line 1124: why do you need this check?

inspire_schemas/utils.py
```diff
@@ -1079,3 +1080,47 @@ def sanitize_html(text):
...
+    schemas = [
+        "hep",
+        "authors",
+        "experiments",
+        "institutions",
+        "conferences",
+        "seminars",
+        "jobs",
+        "journals",
+    ]
```
michamos, line 1104: make this global, so we have a better chance of seeing it when creating a new schema.

tests/unit/test_utils.py
```diff
@@ -1371,3 +1372,81 @@ def test_code_to_name_with_common_name(country_code, expected):
...
+            ("experiments", u"collaboration.record.$ref"),
+            ("experiments", u"new_record.$ref"),
+            ("jobs", u"accelerator_experiments.record.$ref"),
+        ],
+        "related_records": [
+            u"related_records.record.$ref",
+            u"related_records.record.$ref",
+            u"related_records.record.$ref",
+            u"related_records.record.$ref",
+        ],
```
michamos: this doesn't look right. Instead, under `authors` you should have `("authors", "related_records.record.$ref)`, etc.

MJedr: Sorry, I haven't noticed that before, it's already corrected :)

inspire_schemas/utils.py
```diff
@@ -1079,3 +1091,52 @@ def sanitize_html(text):
...
+        Returns:
+            dict(list(tuple)): index and path to the referenced record
+    """
+    references = defaultdict(list)
+    for schema_name in SCHEMAS:
+        schema = load_schema(schema_name=schema_name)
+        for reference_field in get_paths(schema):
+            index_name = reference_field.pop().split(" ")[0]
+            reference_search_path = '.'.join(reference_field)
+            if index_name == "URL":
```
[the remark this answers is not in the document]
MJedr: That is the case for the generic record schema `any.json` - I left the original description, which starts with `URL` and since I am extracting first word from description (because in case of the other reference schemas it is the index name) that is the outcome in case of `related_records`. I can change the description if it would be better.

inspire_schemas/utils.py
```diff
@@ -1079,3 +1091,54 @@ def sanitize_html(text):
...
+            dict(list(tuple)): index and path to the referenced record
+    """
+    references = defaultdict(list)
+    for schema_name in SCHEMAS:
+        schema = load_schema(schema_name=schema_name)
+        for reference_field in get_paths(schema):
+            index_name = reference_field.pop().split(" ")[0]
+            reference_search_path = '.'.join(reference_field)
+            if re.search("deleted_records|self|new_record", reference_search_path):
+                continue
```
michamos, line 1139: I would move this before the joining so you can simply test on `reference_field[0]`.

inspire_schemas/utils.py
```diff
@@ -1079,3 +1091,54 @@ def sanitize_html(text):
...
+            cache[args] = result
+            return result
+    return wrapper
+
+
+def get_paths(schema, previous_node=None):
+    for key, val in schema.items():
+        if isinstance(val, dict):
+            for subkey in get_paths(val, key):
+                if key in ["properties", "items", "description"]:
```
michamos, line 1114: minor: swap these two lines so you're not testing this on every loop iteration.

inspire_schemas/utils.py
```diff
@@ -1079,3 +1091,40 @@ def sanitize_html(text):
...
+    """ For every schema return path and index name for every referenced record
+        Returns:
+            dict(list(tuple)): index and path to the referenced record
+    """
+    if references:
+        return references
+    for schema_name in SCHEMAS:
+        schema = load_schema(schema_name=schema_name)
+        for reference_field in get_paths(schema):
+            if re.search("deleted_records|self|new_record", reference_field[-3]):
```
michamos, line 1122: I don't think you need a regex, which was actually the point:
```suggestion
            if reference_field[0] in {"deleted_records", "self", "new_record"}:
```

inspire_schemas/utils.py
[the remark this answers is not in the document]
michamos: ok, it's probably better to do something like `if reference_field[0] == "related_records"` then, as it makes it more explicit.

inspire_schemas/utils.py
```diff
@@ -1079,3 +1091,40 @@ def sanitize_html(text):
...
+        return references
+    for schema_name in SCHEMAS:
+        schema = load_schema(schema_name=schema_name)
+        for reference_field in get_paths(schema):
+            if reference_field[0] in {"deleted_records", "self", "new_record"}:
+                continue
+            index_name = reference_field.pop().split(" ")[0]
+            reference_search_path = '.'.join(reference_field)
+            if index_name == "URL":
+                references[schema_name].append(("related_records", reference_search_path))
```
michamos, line 1127: This is not correct:
```suggestion
                references[schema_name].append((schema_name, reference_search_path))
```

michamos (approved):
a couple of small things remaining, ok after fixing them :+1:

inspire_schemas/utils.py
MJedr: but we don't know to which index `related_records` are pointing, so in this case we'll obtain something like:
```
{"hep": {"hep": "related_records.records.$ref"}}
```
and I am not sure if it's correct

michamos: that's correct, related records in `hep` point to `hep`, in `institutions` to `institutions`, etc.

PR #416, opened 2020-11-26
github: add actions to replace travis

drjova:


michamos (approved):
:tada: hard to tell if it really works, but let's merge it and try it!

michamos:
Maybe we should disable travis to prevent both deploying to PyPI?

PR #418, opened 2020-12-02
reference_builder: ignore authors without full_name

pazembrz:
* #1553

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -217,17 +217,19 @@ class ReferenceBuilder(object):
             self.add_author(author)
 
     def add_author(self, full_name, role=None):
+        normalized_name = normalize_name(full_name)
+        if not normalized_name:
+            return
         self._ensure_reference_field('authors', [])
-
         if role is not None:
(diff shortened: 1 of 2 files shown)


inspire_schemas/builders/references.py
```diff
@@ -218,17 +218,17 @@ def add_refextract_authors_str(self, authors_str):
...
-            inspire_role = 'editor' if role == 'ed.' else role
-            self.obj['reference']['authors'].append({
-                'full_name': normalize_name(full_name),
-                'inspire_role': inspire_role,
-            })
-        else:
-            self.obj['reference']['authors'].append({
-                'full_name': normalize_name(full_name),
-            })
+        if normalize_name(full_name):
```
michamos, line 221:
```suggestion
        normalized_name = normalize_name(full_name)
        if not normalized_name:
            return
```
so you don't need to indent more and can reuse the variable. Also move it before the `self._ensure_reference_field` as it will create an empty author list if we don't end up adding any authors.

pazembrz: :+1:

michamos (approved):

PR #421, opened 2021-02-23
Add journal_record key to references in HEP

pazembrz:


inspire_schemas/records/elements/reference.yml
```diff
@@ -239,6 +239,8 @@ properties:
                     cited document is a book chapter.
                 minLength: 1
                 type: string
+            journal_record:
+                $ref: json_reference/journal.json
```
michamos, line 243: you should also add it to the record fixture in tests

pazembrz: :+1:

michamos (approved):

PR #422, opened 2021-04-13
utils: add collaboration_name_normalizer

pazembrz:
part of: #1689

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -347,6 +347,8 @@ SCHEMAS = [
     "journals",
 ]
 
+compiled_collaboration_cleaner_regexp = None
+
 
 def _load_countries_data(filename):
     path = resource_filename(__name__, 'countries')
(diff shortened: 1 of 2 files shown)


inspire_schemas/utils.py
```diff
@@ -347,6 +347,8 @@
     "journals",
 ]
 
+compiled_collaboration_cleaner_regexp = None
```
MJedr, line 350: Why do we need this?

pazembrz: It's caching regexp to not "compile" it every time we will call `normalize_collaboration_name`

MJedr (approved):

inspire_schemas/utils.py
michamos: I don't think you ever write to this variable, so I don't think this does what you're claiming.

pazembrz: Ah... You are right! It's my fault... Moving code between repos is not always easy :) Thanks for noticing! I will prepare update in few minutes.

PR #433, opened 2022-03-28
page range: fix bibtex artid return check

nooraangelva:
Ref: https://github.com/inspirehep/inspirehep/issues/2168


drjova:
@nooraangelva could you please write a test?

MJedr (approved):

PR #436, opened 2022-06-13
signature builder: don't raise

drjova:
* Don't raise when the schema is unknown.

* Don't raise when schema don't match the value pattern.

* Ref: cern-sis/issues-inspire/issues/60


MJedr (approved):
Looks good, the only thing I am wondering is that if we won't need to raise this exception at all (on hep for example?)

drjova:
@MJedr which exception?

MJedr:
`SchemaUIDConflict` and `UnknownUIDSchema`, but it's ok, we don't handle it anywhere

drjova:
@michamos could you please double check if this fits what we discussed on Zulip?

michamos (approved):

PR #440, opened 2023-01-31
arxiv: enable doi format in references

MJedr:
ref: cern-sis/issues-inspire#218

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -87,15 +87,36 @@ _RE_AUTHORS_UID = {
 
 # Matches new style arXiv ID, with an old-style class specification
 # (Malformed, but appears in APS records)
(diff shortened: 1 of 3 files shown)


inspire_schemas/utils.py
```diff
@@ -1064,6 +1085,13 @@ def normalize_isbn(isbn):
         return isbn
 
 
+def _get_first_regex_match(regex_list, obj_to_match):
+    for regex in regex_list:
+        match = regex.match(obj_to_match)
+        if match:
+            return match
```
michamos, line 1092: can't you simply join all options with `|` to have one big regex for pre 2007 and one for post 2007?

michamos (approved):
minor comment, but looks ok :+1:

inspire_schemas/utils.py
MJedr: I was thinking about it but I am afraid that it will become unreadable

michamos: I don't mean writing one big string, I like your current approach with the lists. But instead of manually testing them one by one, you could join the regexes in the list with `|` and match them all at once.

MJedr: unfortunately, it can't be done easily cause we define regex groups that can't be redefined in the same regex

PR #461, opened 2024-08-05
global: add pre-commit with ruff

PascalEgn:


drjova (approved):

ruff.toml
```diff
@@ -0,0 +1,30 @@
+target-version = "py311"
+line-length = 100
```
drjova, line 2: can we stick to 88 (black)?

PR #462, opened 2024-08-21
setup: do not use a fixed urllib3 version

miguelgrc:
* ref: https://github.com/cern-sis/issues-inspire/issues/521

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -213,15 +213,12 @@ def do_setup():
             'inspire-idutils==1.2.4; python_version == "2.7"',
             'idutils~=1.2,>=1.2.1; python_version >= "3"',
             'isbnid',
-            'inspire-utils~=3.0,>=3.0.0',
+            'inspire-utils',
             'isodate',
             'pyyaml==5.3.0',
             'pytz',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -213,15 +213,12 @@ def do_setup():
             'inspire-idutils==1.2.4; python_version == "2.7"',
             'idutils~=1.2,>=1.2.1; python_version >= "3"',
             'isbnid',
-            'inspire-utils~=3.0,>=3.0.0',
+            'inspire-utils',
```
PascalEgn, line 216: Could you add `'inspire-utils @ git+https://github.com/miguelgrc/inspire-utils.git@bump-urllib3#egg=inspire-utils',`

Kinda curious what happens then

miguelgrc: I tried and the tests were also passing. The problem comes when also pulling `inspire-schemas` from git in `inspire-utils`

DonHaul:
Fixed in https://github.com/inspirehep/inspire-schemas/commit/d11b04435815acec8404a3deb7da0549fa825097

PR #463, opened 2024-08-22
global: add hepcrawl parsers

miguelgrc:
* ref: https://github.com/cern-sis/issues-inspire/issues/546


setup.py
```diff
@@ -164,12 +164,13 @@ def _generate_json_schemas():
     'isort~=4.0,>=4.3.0',
     'pytest-cache',
     'pytest-cov==2.6.1',
-    'pytest~=3.0,>=3.6.0',
+    'pytest==5.4.3',
```
drjova, line 167: @miguelgrc thanks for the PR, why we need to pin pytest?

miguelgrc: Sorry, in fact we don't need to. I was getting some errors initially with higher versions but after checking again it works with no issues.

drjova (approved):

PR #464, opened 2024-08-23
global: fix parser imports and add missing deps

miguelgrc:
* ref: https://github.com/cern-sis/issues-inspire/issues/546

The `utils.py` file and the `scrapy` and `pylatexenc` dependencies where missing when moving the parsers from `inspire-utils`. This went unnoticed as the tests were still passing since the new version of `inspire-utils` with the parsers removed had not been released yet when the tests ran. It is now released so we should now be able to trust the test results :)


drjova:
Thank you @miguelgrc

PR #467, opened 2024-11-21
Data MVP

michamos:
* Also makes normalization script compatible with recent Python
* Also fixes generation script to run on recent Node versions


inspire_schemas/records/data.yml
```diff
@@ -30,13 +30,334 @@ properties:
...
+                record:
+                    $ref: elements/json_reference/experiment.json
+            type: object
+        minItems: 1
+        title: List of related accelerators/experiments
+        type: array
+        uniqueItems: true
+    acquisition_source:
+        $ref: elements/acquisition_source.json
+    authors:
```
michamos, line 105: In the MVP, this will come from the paper, see also https://github.com/cern-sis/issues-inspire/issues/599, so probably doesn't *need* to be present here (can be done at display/indexing time).

inspire_schemas/records/data.yml
```diff
@@ -88,6 +426,28 @@ properties:
         minLength: 1
         title: Last record update
         type: string
+    literature:
```
michamos, line 429: We will need to link from the data to the paper and back. I've done it like this, comments welcome.

inspire_schemas/records/hep.yml
```diff
@@ -667,6 +667,28 @@ properties:
             are all curated eventually, whereas non-core records are
             not systematically curated.
         type: boolean
+    data:
```
michamos, line 670: I've added this to link from the paper to the data, although it's not strictly needed for the MVP as the link is coming from the data record for HEPData. Should we drop it?

michamos:
I can't fix tests as I can't run them locally because I can't install it due to needing to bump pytest, which blocks due to the `jsonschema2rst` conflict, which I tried to fix in https://github.com/inspirehep/jsonschema2rst/pull/17 which fails because now `py.test` can't be found in the CI :angry:

tests/unit/test_utils.py
```diff
@@ -1424,6 +1424,7 @@ def test_get_references_for_schema_returns_proper_schemas():
             ("jobs", u"accelerator_experiments.record.$ref"),
         ],
         u"hep": [
+            ("data", u"literature.record.$ref"),
```
michamos, line 1427: The test fails because this is not returned, although the schema contains it :confused: it's probably something stupid I don't see

tests/unit/test_utils.py
```diff
@@ -1414,45 +1414,50 @@ def test_code_to_name_respects_overrides():
 
 def test_get_references_for_schema_returns_proper_schemas():
     expected = {
-        u"experiments": [
```
drjova, line 1417: @michamos I had to update these with all the references, could you please check that are correct?

michamos: Thanks @drjova, looks good :+1:

PR #474, opened 2024-12-20
data: add hepdata as acquisitionsource

DonHaul:
* ref: cern-sis/issues-inspire/issues/630

### DIFF
--- a/inspire_schemas/records/elements/acquisition_source.yml
+++ b/inspire_schemas/records/elements/acquisition_source.yml
@@ -55,12 +55,16 @@ properties:
 
             `beard`
                 when created after author disambiguation
+
+            `inspirehep`
+                when obtained through an inspirehep harvest
         enum:
         - submitter
         - oai
         - batchuploader
         - hepcrawl
         - beard
+        - inspirehep
         minLength: 1
         title: How the medata was obtained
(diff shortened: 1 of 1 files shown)


inspire_schemas/records/elements/acquisition_source.yml
```diff
@@ -55,12 +55,16 @@ properties:
...
+
+            `hepdata`
+                when obtained through the HEPData harvest
         enum:
         - submitter
         - oai
         - batchuploader
         - hepcrawl
         - beard
+        - hepdata
```
michamos, line 67: I don't think this should be `hepdata`. The `acquisition_source` is not the same thing as the source, it's meant to represent how we got it (i.e. the tool) rather than the source. So you'd need a name for the new harvesting component.

DonHaul: I see, to be checked then what should be its name, maybe something like `backofficeharvester`

drjova: let's call it `inspirehep`, wdyt?

DonHaul: sounds good

PR #476, opened 2025-01-14
feat[DataBuilder]: Pass doi value directly when adding literature

Lorenzovagliano:
- Also Added record validation in unit tests in order to avoid testing wrong behavior, as was previously happening with test_add_literature.

ref: https://github.com/cern-sis/issues-inspire/issues/631

### DIFF
--- a/inspire_schemas/builders/data.py
+++ b/inspire_schemas/builders/data.py
@@ -297,20 +297,26 @@ class DataBuilder(RecordBuilder):
         self._append_to('urls', entry)
 
     @filter_empty_parameters
-    def add_literature(self, doi=None, record=None):
+    def add_literature(self, doi=None, record=None, source=None):
         """Add literature.
 
         :param doi: doi of the literature
-        :type doi: dict
+        :type doi: str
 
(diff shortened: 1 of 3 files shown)


inspire_schemas/builders/data.py
PascalEgn, line 1: With these changes we are no longer able to add a source to the DOI of a Literature record (but it should be possible given the Schema https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/data.yml#L442-L443).

Maybe you can pass something like `doi_value` and `doi_source`.

But to be discussed if this is even really needed...

Lorenzovagliano: The way we are currently using it on inspirehep in `workflows/dags/data/data_harvest.py` is: 

` builder.add_literature(doi={"value": doi}, record={"$ref": `f"{inspire_url}/api/literature/{inspire_id}"},`

So the doi_source is not being used, but maybe that could be improved. Let's see if it's needed.

Lorenzovagliano: Added source parameter.

PascalEgn (approved):

PR #482, opened 2025-04-14
Masterpy2py3

DonHaul:


scripts/version_manager.py
```diff
@@ -543,33 +551,41 @@ def get_releasenotes(repo_path, from_commit):
...
-        version_str,
-        '\n'.join(reversed(api_break_changes)),
-        '\n'.join(reversed(features)),
-        '\n'.join(reversed(bugs)),
-    )
+            version_str,
+            '\n'.join(reversed(api_break_changes)),
+            '\n'.join(reversed(features)),
+            '\n'.join(reversed(bugs)),
+        )
```
drjova, line 558: why we have tabs?

setup.py
```diff
@@ -226,15 +212,33 @@ def do_setup():
...
+        ],
         tests_require=tests_require,
         extras_require=extras_require,
         build_requires=build_requires,
         license='GPLv2',
         name='inspire-schemas',
         package_data={'': ['*.json', 'CHANGELOG', 'AUTHORS']},
         packages=find_packages(),
-        setup_requires=['pyyaml>=6.0,<7.0'],
+        setup_requires=['pyyaml==5.3.0'],
```
drjova, line 241: don't we have to pin it as in `build`?

drjova (changes requested):
Looks good, few comments

.github/workflows/build-and-release.yml
```diff
@@ -8,12 +8,10 @@ on:
...
 jobs:
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        include:
-          - python: python3
-            pip: pip3
+        python: ['3.11']
```
drjova, line 14: I would include 12, 13 as optional

drjova (changes requested):
Thanks @DonHaul looks really good. Just few comments

scripts/version_manager.py
```diff
@@ -43,7 +42,9 @@
 import dulwich.walk
 
 PROJECT_NAME = 'inspire-schemas'
-BUG_URL_REG = re.compile(r'.*(closes #|fixes #|adresses #)(?P<bugid>\d+)')
+BUG_URL_REG = re.compile(
```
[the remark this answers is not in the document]
DonHaul: indeed it isn't it was previously being used to create releases. removed

.github/workflows/build-and-release.yml
DonHaul: since they are passing. Added them as required

drjova (approved):
LGTML, please just make sure it generates correctly the json files and releases on npm too

PR #484, opened 2025-05-06
isbn: use new isbnlib

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/902

### DIFF
--- a/inspire_schemas/builders/references.py
+++ b/inspire_schemas/builders/references.py
@@ -27,11 +27,12 @@ from __future__ import absolute_import, division, print_function
 import re
 
 import idutils
+import isbnlib
 import six
 from inspire_utils.date import normalize_date
 from inspire_utils.name import normalize_name
 from inspire_utils.record import get_value
-from isbn import ISBN
+from inspire_utils.isbn import normalize_isbn
 
 from inspire_schemas.utils import (
     convert_old_publication_info_to_new,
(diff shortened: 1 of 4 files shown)


inspire_schemas/builders/references.py
```diff
@@ -321,12 +322,11 @@ def _add_uid(self, uid, skip_handle=False):
                 }
             )
         else:
-            # ``idutils.is_isbn`` is too strict in what it accepts.
-            try:
-                isbn = str(ISBN(uid))
+            if isbnlib.is_isbn10(uid) or isbnlib.is_isbn13(uid):
```
drjova, line 325: do we need the check? isn't already in normilize?

drjova (changes requested):

inspire_schemas/builders/references.py
ioannistsanaktsidis: we need to check also here. Normalize is catching the exception [silently ](https://github.com/inspirehep/inspire-utils/commit/b10cdbdd95003934b7b5b4d9eccbfde3c6670bb6#diff-dcf6904ef411b4f38b1aa178da22c25b55da1376fea25bfdf7f884bbae877c7cR35)(so as to ensure minimal changes in the repos and keep the same behavior on the normalize as before). 

We have cases like this one here https://github.com/inspirehep/inspire-schemas/blob/77767929a03207f58327cf0a850ee93c2ecb9303/tests/unit/test_reference_builder.py#L1089-L1107 . This will fail as `normalize` tries to normalize the isbn and if it cannot, it returns what was passed. So in that case it will put the `cds` link in the isbn, which is wrong. 

Thats why I am preventing it like this, otherwise, we will need to write a very similar logic as the `normalize` one and just  let it raise here

drjova (approved):

PR #495, opened 2025-10-06
jsonschema: use fast validator for py3

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/938

### DIFF
--- a/inspire_schemas/utils.py
+++ b/inspire_schemas/utils.py
@@ -28,6 +28,7 @@ import copy
 import json
 import os
 import re
+import sys
 from collections import defaultdict
 from functools import partial, wraps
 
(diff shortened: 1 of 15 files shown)


inspire_schemas/utils.py
```diff
@@ -741,7 +748,30 @@ def load_schema(schema_name, resolved=False, _cache={}):  # noqa: B006
...
 
 
-inspire_format_checker = draft4_format_checker
+def _make_hashable(item):
+    if isinstance(item, (str, int, float, bool, type(None))):
+        return item
+    return json.dumps(item, sort_keys=True, separators=(",", ":"))
+
+
+def fast_unique(validator, uivalue, instance, schema):
```
drjova, line 757: `uivalue`? typo? Shouldn't be value?

inspire_schemas/utils.py
```diff
@@ -741,7 +748,30 @@ def load_schema(schema_name, resolved=False, _cache={}):  # noqa: B006
...
+    if not uivalue or not isinstance(instance, list):
+        return
+    seen = set()
+    for item in instance:
+        key = _make_hashable(item)
+        if key in seen:
+            yield ValidationError("%r has non-unique elements" % (instance,))
+            return
+        seen.add(key)
```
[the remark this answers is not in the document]
ioannistsanaktsidis: seen is used to hold the hashable so as to detect if there is a duplicate or not .

inspire_schemas/utils.py
drjova: yeah it's confusing, let's call it `unique_items_value` or similar, to make sure it's clear:)

